PR #2, opened 2023-04-13
Bifrost changes

### LINKED ISSUE #1: Indexer Integration for Zenlink
As part of our integration with Zenlink we want to properly define the indexer

# To Do
- Inspect the bifrost-kusama indexer
- Iteratively remove elements that are not related to our chain or zenlink
- Test that it works as expected

# Resources
- https://github.com/pendulum-chain/squid-amplitude
- https://github.com/zenlinkpro/bifrost-kusama-squid
# Current Status
Researching the bifrost-kusama repository

LeoFraq:
This PR is an attempt at generating all the zenlink necessary classes from the [bifrost](https://github.com/zenlinkpro/bifrost-kusama-squid) example. 

This has been done iteratively so as to create only the (supposedly) necessary files and modules.

Do note that you will need to regenerate everything under lib/ and src/model/generated as by the readme

- The bifrost schema contains a lot of types that we do not need/have, we can prune that and the whole thing should be a lot cleaner
- The code runs, however, previously we had only one Event defined (Balance.Transfer), now we have a lot of types and we should probably define events for those too?
- I switched the archive node of reference in both the processor.ts and typegen.json
- It was using kusama as a reference which I believe was not intended? We want the schema to be the one from amplitude

Besides that, I'd say the relevant parts are all under
- processor.ts
- types/events.ts
- mappings/protocol.ts
Everything else seems to go and support methods/classes used in these files, so we could theoretically write our own versions of those supporting files


src/config.ts
```diff
@@ -0,0 +1,10 @@
+import { ProcessorConfig } from "./types";
+
+export const config: ProcessorConfig = {
```
LeoFraq: Verify the data here is correct

src/constants.ts
```diff
@@ -0,0 +1,36 @@
+import { TokenBase } from "./types";
+import { Big as BigDecimal } from 'big.js'
+import { AssetId } from "./types/v906";
+import { zenlinkAssetIdToCurrencyId } from "./utils/token";
+
+// Probably need our own tokens
```
LeoFraq, line 6: Add every token we are planning on using on the swap, remove comment

src/processor.ts
```diff
@@ -32,6 +38,8 @@ type Item = BatchProcessorItem<typeof processor>
 type Ctx = BatchContext<Store, Item>
 
 
+// TODO implement handleXXX as Bifrost
```
LeoFraq: See the bifrost implementation for this, relevant once the code actually runs

src/types/events.ts
```diff
@@ -59,3 +60,143 @@ export class BalancesTransferEvent {
         return this._chain.decodeEvent(this.event)
     }
 }
+
```
LeoFraq: The bifrost sample contains A LOT more events, this has been trimmed and only contains the ones that seem to be necessary for the current iteration. We can add as necessary all the others

src/types/events.ts
```diff
@@ -59,3 +60,143 @@ export class BalancesTransferEvent {
...
+        event = event || ctx.event
+        assert(event.name === 'ZenlinkProtocol.AssetSwap')
+        this._chain = ctx._chain
+        this.event = event
+    }
+
+    /**
+     *  Transact in trading \[owner, recipient, swap_path, balances\]
+     */
+    get isV902(): boolean {
```
LeoFraq: Do note: the bifrost implementation has more isVXXX/ getVXXX.  I believe we will need fewer implementations of this as we trade less tokens. Trimmed appropriately.

src/processor.ts
```diff
@@ -1,20 +1,22 @@
...
-        // Lookup archive by the network name in the Subsquid registry
-        //archive: lookupArchive("kusama", {release: "FireSquid"})
+const DataSelection = { data: { event: true } } as const
 
-        // Use archive created by archive/docker-compose.yml
-        archive: lookupArchive('kusama', {release: 'FireSquid'} )
-    })
+
+const processor = new SubstrateBatchProcessor()
+    .setDataSource(config.dataSource)
```
LeoFraq, line 23: why were we still using kusama as a data source? Switched to amplitude archive node

src/processor.ts
```diff
@@ -82,35 +84,35 @@ interface TransferEvent {
...
+                case 'ZenlinkProtocol.LiquidityAdded':
+                    await handleLiquidityAdded({ ...ctx, block: block.header, event: item.event })
+                    break
+                case 'ZenlinkProtocol.LiquidityRemoved':
+                    await handleLiquidityRemoved({ ...ctx, block: block.header, event: item.event })
+                    break
+                case 'ZenlinkProtocol.AssetSwap':
+                    await handleAssetSwap({ ...ctx, block: block.header, event: item.event })
+                    break
+                case 'Balances.Transfer':
```
LeoFraq, line 86: Unsure what this casing actually does, we can ask nikita for this

typegen.json
```diff
@@ -1,11 +1,17 @@
 {
   "outDir": "src/types",
-  "specVersions": "https://kusama.archive.subsquid.io/graphql",
+  "specVersions": "https://amplitude.archive.subsquid.io/graphql",
```
LeoFraq: Again: why were we referring to the kusama archive and not the amplitude?

ebma: @RustNinja do you remember something about this? Is it just an oversight or was there a reason?

LeoFraq:
As of latest commit the code builds and runs "correctly"

ebma:
I re-added some files because there is no need to hide those. However I can't run it locally.
When I follow the steps in the README, `sqd up` fails for me with 
```
$ sqd up
UP
    Error: spawn docker-compose ENOENT
    Code: ENOENT

```

Note that I also had to install `sqd` as a global package with npm because the `brew` build was corrupt somehow.

@LeoFraq are you able to get it running with the steps in "Getting started"?

LeoFraq:
@ebma  I actually used the instructions in bifrost's example which worked for me.
not sure if they changed between @RustNinja 's changes or if theres any difference

ebma:
So all of these commands
```
# 1. Install dependencies
npm ci

# 2. Compile typescript files
make build

# 3. Start target Postgres database and detach
make up

# 4. Start the processor
make process

# 5. The command above will block the terminal
#    being busy with fetching the chain data, 
#    transforming and storing it in the target database.
#
#    To start the graphql server open the separate terminal
#    and run
make serve
```

worked for you @LeoFraq ?

ebma:
I started re-adding also the code/events related to bifrosts farming pallet as we recently decided to use that pallet for our AMM LP rewards. Still WIP though

ebma:
I want to get this PR and branch out of the way. Unfortunately, it became quite large and also contains changes unrelated to Bifrost. I will merge these changes to `main` so that we can start anew and create a cleaner process for this repo next.

PR #6, opened 2023-07-10
4 finish and test zenlink implementation

### LINKED ISSUE #4: Finish and test Zenlink implementation
Although we already added some changes to our indexers to support Zenlink, we never actually tried if it works in practice because we were not really able to interact with the Zenlink pallet yet.

## TODO 
- [x] Create a pool/pair on Foucoco's Zenlink pallet
- [ ] Check that the related events are properly processed by the indexer. If that's not the case, add the required changes

adelarja:
This PR aims to complete the squid configuration so that the Zenlink Protocol events can be processed.


src/types/amplitude/v8.ts
```diff
@@ -16,6 +20,11 @@ export interface CurrencyId_Stellar {
     value: Asset
 }
 
+export interface CurrencyId_ZenlinkLPToken {
```
adelarja, line 23: I manually added this because, for some reason, it is not generated from the archive using sqd typegen.

ebma: For Amplitude I think our ZenlinkLPToken change is just not live yet.

ebma (commented):
I think I'm overall happy with this now. 
@adelarja let me know if all events/data are indeed queryable as expected from the graphql endpoint and I think we can merge this afterwards.

PR #8, opened 2023-08-04
Add changes for indexing events of Nabla smart contracts

### LINKED ISSUE #7: Add changes for indexing events of Nabla smart contracts
This ticket is related to https://github.com/pendulum-chain/tasks/issues/16.

We want to make sophisticated UIs for interaction with the Nabla smart contracts. In order to do so, we need to collect and transform some historic data emitted my the Nabla smart contracts.
This ticket is about adding these changes to our indexers.

The implementation should draw heavy inspiration from [this](https://github.com/0xamberhq/subgraph-mumbai) existing implementation of an indexer for Nabla.

adelarja:
This ticket is related to [this](https://github.com/pendulum-chain/pendulum-squids/issues/7) issue and [pendulum-chain/tasks/issues/16](https://github.com/pendulum-chain/tasks/issues/16).

We need to collect and transform some historic data emitted my the Nabla smart contracts.
The aim of this PR is to add these changes to our indexers.


ebma (commented):
Just some preliminary feedback, I know it's just a draft PR. 

I suppose the contents of the `abis/*.json` and `src/abi/*.ts` files are the same but in the `.ts` files you make it easier to import the data? Before we merge we probably should remove the files that are not required by the indexer, and I assume the contents of the root `abis` folder is not really required (?)

Also, I see that right now in `src/mappings/nabla.ts` we define hard-coded addresses of all the contracts that we expect and want to watch over. Can we change the code so that the indexer picks up these events on all smart contracts? Because during the development we will probably have to deploy a lot of different instances of `Router`s, `BackstopPool`s, `SwapPool`s. And it does not make sense to use all of these instances at once, eventually once Nabla 'goes live' we'd of course only use one set of these contract instances. But I think choosing which set of instances should be done in the UI that interacts with the contracts. We should not make this assumption in the indexer and thus I'd hope we can provide events for any Nabla smart contract that is deployed. Without having to hard-code any address in the indexer. Do you think this is possible @adelarja?

adelarja:
> Just some preliminary feedback, I know it's just a draft PR.
> 
> I suppose the contents of the `abis/*.json` and `src/abi/*.ts` files are the same but in the `.ts` files you make it easier to import the data? Before we merge we probably should remove the files that are not required by the indexer, and I assume the contents of the root `abis` folder is not really required (?)

@ebma Many thanks for the comment. I've deleted all the files in the abis folder. I was thinking that maybe we could leave the .json files in order to know which contract version we are using (having the abi and metadata as evidence), but it is not mandatory at all. As you said, it is only used for generating the ts files with the needed functions to handle the events.

> Also, I see that right now in `src/mappings/nabla.ts` we define hard-coded addresses of all the contracts that we expect and want to watch over. Can we change the code so that the indexer picks up these events on all smart contracts? Because during the development we will probably have to deploy a lot of different instances of `Router`s, `BackstopPool`s, `SwapPool`s. And it does not make sense to use all of these instances at once, eventually once Nabla 'goes live' we'd of course only use one set of these contract instances. But I think choosing which set of instances should be done in the UI that interacts with the contracts. We should not make this assumption in the indexer and thus I'd hope we can provide events for any Nabla smart contract that is deployed. Without having to hard-code any address in the indexer. Do you think this is possible @adelarja?

I was able to handle this situation.

I've used some nested try catch statements to solve the issue in a more generic way. I'll push it today. This is basically what I did:

```typescript
enum EventType {
    BackstopPoolEvent,
    RouterEvent,
    SwapPoolEvent
}

function getEventAndEventType(ctx: EventHandlerContext) {
    
    try {
        const event = bpool.decodeEvent(ctx.event.args.data);
        const eventType = EventType.BackstopPoolEvent;
        return {event, eventType}
    } catch {
        try {
            const event = spool.decodeEvent(ctx.event.args.data);
            const eventType = EventType.SwapPoolEvent;
            return {event, eventType};
        } catch {
            try {
                const event = rou.decodeEvent(ctx.event.args.data);
                const eventType = EventType.RouterEvent;
                return {event, eventType};
            } catch {
                const event = undefined;
                const eventType = undefined;
                return { event, eventType };
            }
        }
    }
```

Depending on the eventType we can use a particular function to create/modify/update the database.

@ebma **There is one thing to note here: For some reason, the Token Contracts ( mUSD deployed to 6h6JMHYBV7P6uQekZXzHmmpzx7tzHutTyx448MnFogR6dNde,  mEUR deployed to 6hJHAXrCYNFbuDzgwZb2S1UhamDBtR8Jw1cn9gKQ4QewvSh1 and  mETH deployed to 6kwiNhGTHAfGb5x3gZBQxhiG2rf9F8W7Da3HhQRdBHQopSHv) are not generating errors when trying to decode them as BackstopPool events. That is reflected as false backstop pools being generated from our squid. I'll try to find a way to solve this. This is the main disadvantage of the generic implementation.**

adelarja:
@ebma I added a way to handle the events in a more generic way using the function named [getEventAndEventType](https://github.com/pendulum-chain/pendulum-squids/blob/f5ad37a69fbb769c3c5c6a2b62a48651a144f522/src/mappings/nabla.ts#L52C1-L74C2) is used.

However, I left the [old code](https://github.com/pendulum-chain/pendulum-squids/blob/f5ad37a69fbb769c3c5c6a2b62a48651a144f522/src/constants.ts#L49C1-L53C2) related to the hard-coded smart contract addresses. In case we want to use it in the future, we could add all the backstop pools, swap pools and routers addresses into the [foucoco](https://github.com/pendulum-chain/pendulum-squids/blob/f5ad37a69fbb769c3c5c6a2b62a48651a144f522/src/constants.ts#L39C1-L47C2) or [amplitude](https://github.com/pendulum-chain/pendulum-squids/blob/f5ad37a69fbb769c3c5c6a2b62a48651a144f522/src/constants.ts#L49C1-L53C2) constants. I think it could be valuable fpr more efficient event management if needed.

I also left the old implementation commented for you to compare. as shown [here](https://github.com/pendulum-chain/pendulum-squids/blob/f5ad37a69fbb769c3c5c6a2b62a48651a144f522/src/constants.ts#L49C1-L53C2).

ebma:
Thanks for the update @adelarja. Did you manage to find a solution for the problem of ERC20 contract events being decoded to BackstopPool events? Or is that still a pending issue?

adelarja:
> Thanks for the update @adelarja. Did you manage to find a solution for the problem of ERC20 contract events being decoded to BackstopPool events? Or is that still a pending issue?

@ebma The solution I found is to try to decode the event as an ERC20 token event. In case it is, we simply skip the event.

I thought that the ERC20 tokens were being registered as BackstopPools, but then I decoded the addresses of the backstop pools, and it seems they are different contracts. I didn't know that we have 4 backstop pool contracts deployed on our network, haha.

Now, everything seems to be working well.

schema.graphql
```diff
@@ -692,3 +692,40 @@ type Incentive @entity {
...
+# Nabla entities
+
+type Router @entity {
+    id: ID! # id can be the address of the router contract
+    swapPools: [SwapPool!] @derivedFrom(field: "router")
+    backstopPools: [BackstopPool!] @derivedFrom(field: "router")
+    paused: Boolean! # whether the router is paused
+}
+
+type NablaToken @entity {
```
adelarja, line 705: I had to rename from Token to NablaToken, since we already had a Token entity in the schema.

ebma: Nice, that makes sense. We just have to keep it in mind for the portal queries.

ebma (approved):
I added some minor refactorings and changed it such that the abi of the 'real' ERC20 is used and not of the mock contract. It still works the same, I tested it. 

Great job overall @adelarja 👍 , I'm very happy that we figured out a way to make this work without hard-coded contract addresses. 
This is ready for merge IMO. It would be nice if you could add a very brief description to the PR though. 1-2 sentences are enough.

PR #10, opened 2023-08-10
9 add processing for farming events to indexer

### LINKED ISSUE #9: Add processing for 'farming' events to indexer
As part of #4, we implemented the processing of basic AMM events. Since we want to use Bifrost's farming pallet for generating rewards for LP token holders, we also need to add the processing for events emitted by that pallet.

## TODO
Look at https://github.com/zenlinkpro/bifrost-kusama-squid/tree/main and adopt all logic that is required for processing the 'farming'-related events and eventually expose the data similar to the bifrost indexer.

ebma:
This PR adds support for farming events, and also processing of our on-chain price updates to the squid. The price data is needed to derive and provide prices for our pool tokens.

## TODO
- [x]  ~Remove stored farming pool data when killed~. I decided not to change this, because the killing of a farming pool is handled very thoroughly in the Bifrost indexer, see [here](https://github.com/zenlinkpro/bifrost-kusama-squid/blob/d51eca107800c8f8be13400c76e4eac1cc941087/src/utils/farming.ts#L395), and I don't want to remove this logic without knowing all the implications.
- [x] Add proper event handling for the Amplitude squid. Currently only the foucoco events are considered.

## Notes
The naming of the functions in `utils/pricing.ts` is a bit confusing but I left it the way it is in the bifrost squid. As far as I understand, our native currency is represents 'eth', thus the return value of `findEthPerToken()` is 1 unit for our native currency. The prices for other assets are then represented in relation to our native currency amount (aka. 'eth').
The derivation of the price of our native asset now happens based on the AMPE/KSM pair. We have access to the real-world KSM price since it's fed to our on-chain storage by the off-chain worker every block. So to derive the price of our native token we check the ratio of the pair balances to get the value of the native asset compared to KSM, and then check that value to the KSM price.

Closes #9.


src/constants.ts
```diff
@@ -1,34 +1,28 @@
...
-import { zenlinkAssetIdToCurrencyId } from './utils/token'
 
-// Probably need our own tokens
-export const TOKEN_METADATA_MAP: { [address: string]: TokenBase } = {
-    '2124-0-0': { name: 'AMPLITUDE', symbol: 'AMPE', decimals: 12 },
-    '2124-2-256': { name: 'Kusama', symbol: 'KSM', decimals: 12 },
-    '2124-2-512': { name: 'Stellar Native', symbol: 'XLM', decimals: 12 },
-    '2124-2-513': { name: 'Stellar USDC', symbol: 'USDC', decimals: 12 },
-    '2124-2-514': { name: 'Stellar TZS', symbol: 'TZS', decimals: 12 },
-    '2124-2-515': { name: 'Stellar BRL', symbol: 'BRL', decimals: 12 },
```
adelarja, line 13: Why were USDC, TZS, and BRL deleted?

A few days ago, I made some tests by creating a pool with Stellar BRL and XCM(0) tokens (only for testing purposes), and it was visible in the token query. However, it seems that it's not working anymore.

adelarja (dismissed):
Overall, it looks good to me. I only have a doubt about deleting the Stellar assets from the TOKEN_METADATA_MAP. I mean, we might not need it in the near future, but it could be a good idea to consider keeping it in case we want to use it later.

src/constants.ts
ebma: I removed them because only AMPE and KSM are really needed. However, you have a point in that we can also expose the data for the rest, it would not do any harm. I will re-add them.

ebma: I re-added the other currencies now. Please have another look @adelarja

adelarja (approved):

PR #12, opened 2023-08-24
11 fix contract ids not showing encoded addresses

### LINKED ISSUE #11: Fix contract `id`s not showing encoded addresses
At the moment, the `id`s that we assign to the nabla smart contracts are shown as hex-encoded values.

image

 However, we'd like to show 'proper' addresses like `6j8TG1juRth1aWBrAh1PvgB7FTmfnFkrTHgg83vjzRnGiWnQ` as this parameter. 

## TODO
Change the assigned `id` parameter to an ss58 encoded address. 
I assume this can be achieved with the `codec.encode()` function, similar to [here](https://github.com/pendulum-chain/pendulum-squids/blob/b28c0c9e90bd4b379e45effcabe4c0ee63ad60dc/src/mappings/balances.ts#L11-L12).

adelarja:
The addresses in the router, swap pool, and backstop pool IDs were saved as hex addresses. However, for the Nabla project, we require SS58 addresses.

This pull request modifies how we store addresses in the entities. Additionally, we identified an error when decoding events, which has been resolved by making calls to the smart contracts to determine the origin of the event.

Closes [this](https://github.com/pendulum-chain/pendulum-squids/issues/11) issue.

### DIFF
--- a/src/mappings/nabla.ts
+++ b/src/mappings/nabla.ts
@@ -87,7 +87,7 @@ async function getEventAndEventType(ctx: EventHandlerContext): Promise<{
     event: Event | null
     eventType: EventType | null
 }> {
-    const veryfiers = [isBackstopPoolEvent, isRouterEvent, isSwapPoolEvent]
+    const verifiers = [isBackstopPoolEvent, isRouterEvent, isSwapPoolEvent]
     const decoders = [backstopPoolAbi, routerAbi, swapPoolAbi]
     const eventTypes = [
         EventType.BackstopPoolEvent,
--- a/src/mappings/nabla.ts
+++ b/src/mappings/nabla.ts
@@ -30,18 +30,6 @@ function ss58ToHex(address: string) {
     return toHex(ss58.decode(address).bytes)
 }
 
-function decodeEvent(
-    data: any,
-    decoder: Decoder
-): { result: Event | null; error: Error | null } {
-    try {
-        const event = decoder.decodeEvent(data)
-        return { result: event, error: null }
-    } catch (err: any) {
-        return { result: null, error: err }
-    }
-}
-
 async function isSwapPoolEvent(ctx: EventHandlerContext) {
     try {
         const contract = new swapPoolAbi.Contract(ctx, ctx.event.args.contract)
@@ -49,9 +37,9 @@ async function isSwapPoolEvent(ctx: EventHandlerContext) {
         await contract.backstop()
         await contract.accumulatedSlippage()
         await contract.poolCap()
-        return true
+        return swapPoolAbi.decodeEvent(ctx.event.args.data)
     } catch {
-        return false
+        return undefined
     }
 }
 
@@ -59,9 +47,9 @@ async function isRouterEvent(ctx: EventHandlerContext) {
     try {
         const contract = new routerAbi.Contract(ctx, ctx.event.args.contract)
         await contract.poolByAsset(new Uint8Array(32))
-        return true
+        return routerAbi.decodeEvent(ctx.event.args.data)
     } catch {
-        return false
+        return undefined
     }
 }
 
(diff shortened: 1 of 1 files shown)


src/mappings/nabla.ts
```diff
@@ -34,35 +42,77 @@ function decodeEvent(
...
+
 // Iterates over all decoders and returns the first successfully decoded event
-function getEventAndEventType(ctx: EventHandlerContext): {
+async function getEventAndEventType(ctx: EventHandlerContext): Promise<{
     event: Event | null
     eventType: EventType | null
-} {
-    const decoders = [erc20Abi, backstopPoolAbi, swapPoolAbi, routerAbi]
+}> {
+    const veryfiers = [isBackstopPoolEvent, isRouterEvent, isSwapPoolEvent]
```
ebma: Nit: typo
```suggestion
    const verifiers = [isBackstopPoolEvent, isRouterEvent, isSwapPoolEvent]
```

src/mappings/nabla.ts
```diff
@@ -34,35 +42,77 @@ function decodeEvent(
...
+        )
+        await contract.getBackedPool(0n)
+        return true
+    } catch {
+        return false
+    }
+}
+
+async function verifyEvent(veryfier: Function, ctx: EventHandlerContext) {
+    return veryfier(ctx)
```
ebma:
```suggestion
async function verifyEvent(verifier: Function, ctx: EventHandlerContext) {
    return verifier(ctx)
```

src/mappings/nabla.ts
```diff
@@ -34,35 +42,77 @@ function decodeEvent(
...
-    for (let i = 0; i < decoders.length; i++) {
-        const { result: event, error: err } = decodeEvent(
-            ctx.event.args.data,
-            decoders[i]
-        )
-        if (!err) {
-            const eventType = eventTypes[i]
-            return { event, eventType }
+    for (let i = 0; i < veryfiers.length; i++) {
+        if (await verifyEvent(veryfiers[i], ctx)) {
```
ebma: Nit: We can simplify this logic now. The implementations of the `verifiers` that are called as part of the `verifyEvent` function do decode the contracts. Thus, the call to `decodeEvent()` in ll. 100-102 is actually redundant _if_ we pass the decoded event out of the verifiers. 

We could change the return type of the verifier functions to the event instead of a boolean. And in l. 99 we then check if the returned `event !== undefined`. However, changing it this way might not necessarily make the code easier to read/understand.

Maybe it's enough if we just add a comment here. WDYT @adelarja?
```suggestion
    // Iterate over all verifiers and try to decode the event to the given type
    for (let i = 0; i < verifiers.length; i++) {
        if (await verifyEvent(verifiers[i], ctx)) {
```

adelarja: > Nit: We can simplify this logic now. The implementations of the `verifiers` that are called as part of the `verifyEvent` function do decode the contracts. Thus, the call to `decodeEvent()` in ll. 100-102 is actually redundant _if_ we pass the decoded event out of the verifiers.
> 
> We could change the return type of the verifier functions to the event instead of a boolean. And in l. 99 we then check if the returned `event !== undefined`. However, changing it this way might not necessarily make the code easier to read/understand.
> 
> Maybe it's enough if we just add a comment here. WDYT @adelarja?

@ebma thanks for the comment.

I totally agree with you, we don't need to call decodeEvent once we know the event type.

While I was improving the logic, something came to my mind:

If in the future the contracts are upgraded, and for example, new events are emitted, or the function names change, or whatever. In that case we would have two choices:
1. Track only events emitted by the upgraded smart contracts, updating the ABIs.
2. Track both, the events emitted by the upgraded contracts and by the old smart contracts (using both the old and the updated ABIs).

It is not urgent, but it might be good to keep it in mind for the future.

ebma (approved):
LGTM, good job 👍

src/mappings/nabla.ts
ebma: That's a good point but I think practically speaking we will not face this issue in the future. The reason being that smart contracts are not upgradable, at least not without putting in some extra work, which AFAIK Nabla is not planning to add. So once Nabla launches on Pendulum, the smart contracts will probably (or ideally) stay like that forever and we don't have to worry about ABIs changing over time.

PR #21, opened 2023-10-26
19 subscription feature for mykobo to subscribe to the pendulum indexer for transaction tracking on pendulum

### LINKED ISSUE #19: Subscription feature for Mykobo to subscribe to the Pendulum Indexer for transaction tracking on Pendulum
Develop a subscription feature that allows Mykobo to subscribe to the Pendulum Indexer so the transactions on Pendulum can be tracked

Acceptance Criteria:
The Pendulum Squid is deployed to Aquarium and accessible
The subscription feature for external parties is implemented
Document the subscription API or mechanism for external parties.

## Technical considerations
The implementation should be done using the GraphQL subscriptions, which are documented for subsquid [here](https://docs.subsquid.io/graphql-api/subscriptions/).

adelarja:
Issue #19 

This PR adds the `TokenTransfer` entity to our squid schema. It also changes some configuration commands that allow us to expose a subscription endpoint.

The idea behind this new feature is to make it possible to subscribe to our squid and detect transactions using a specific endpoint.

~PS: I still need to add the handling of `Deposit` and `Withdraw` events (pretty much the same).~


src/mappings/token.ts
```diff
@@ -59,7 +62,22 @@ export async function handleTokenDeposited(ctx: EventHandlerContext) {
...
+    if (!event) return
+
+    const tokenDeposit = new TokenDeposit({
+        id: ctx.event.id,
+        blockNumber: ctx.block.height,
+        timestamp: new Date(ctx.block.timestamp),
+        extrinsicHash: ctx.event.extrinsic?.hash,
+        who: codec(config.prefix).encode(event.who),
+        amount: event.amount,
+        currencyId: JSON.stringify(event.currencyId),
```
ebma: Let's maybe create a new function that converts a `CurrencyId` to a string? The output of `JSON.stringify()` looks a bit ugly.
How about we make it such that the function spits out more readable strings like
`Native`, `XCM(0)`, `Stellar::StellarNative`, `Stellar::AlphaNum4(issuer,code)`, `Stellar::AlphaNum12(issuer,code)`, and `ZenlinkLpToken(x,x,x,x)` ?

ebma (commented):
Logic looks good overall :+1: 
but I did not try locally yet

src/mappings/token.ts
adelarja: I totally agree, @ebma . I've made the changes you suggested.

scripts/suscribeTransfer.js
```diff
@@ -0,0 +1,87 @@
+const WebSocket = require('ws')
```
adelarja: With this example snippet, we are able to subscribe to a ws endpoint and detect when new `tokenTransfers`, `tokenDeposits` and `tokenWithdrawns` occur.

cc: @ebma

src/mappings/token.ts
```diff
@@ -59,7 +62,71 @@ export async function handleTokenDeposited(ctx: EventHandlerContext) {
...
+                }
+                case 'object': {
+                    currencyId =
+                        'XCM(' + String(event.currencyId.value.__kind) + ')'
+                    break
+                }
+            }
+            break
+        }
+    }
```
ebma: Let's put this logic into an extra function and re-use that function to deduplicate code.
Maybe call it `function currencyIdToString(currencyId: <the type>)` and then pass the `event.currencyId` into it. Not sure what `event.currencyId` has as a type but in case of doubt we can also use `any`.

scripts/suscribeTransfer.js
```diff
@@ -0,0 +1,102 @@
...
+subscription {
+  tokenTransfers(where: {from_eq: "6mYqNJsaGLNduqMdBFhj3ZdzkLy9kghThEUHE3MswphpoATv", to_eq: "6izw2Zx6zcgA44G51ptCKWbsLeyDE4hgVcvKifAkfDr2tATc", currencyId_eq: "XCM(0)"}) {
+    from
+    to
+    amount
+    currencyId
+  }
+}
+`
+*/
```
ebma: Can we also have an example subscription that filters for a specific`currencyId` like you have here in your query. Let's maybe create a sample subscription to listen to XCM(0) transfers

scripts/suscribeTransfer.js
ebma: Nice, that's amazing 🎉

src/mappings/token.ts
```diff
@@ -32,6 +35,59 @@ async function isCompleteMint(
...
+        case 'Stellar': {
+            switch (event.currencyId.value.__kind) {
+                case 'StellarNative': {
+                    currencyId = 'StellarNative'
+                    break
+                }
+                case 'AlphaNum4': {
+                    currencyId =
+                        'Stellar::AlphaNum4(' +
+                        String(event.currencyId.value.code) +
```
ebma, line 67: Oh and can we also have the `issuer` field here and for `AlphaNum12`?
Something like 
```
Stellar::AlphaNum4(' +
                        String(event.currencyId.value.code) 
                        + "," 
                        + String(event.currencyId.value.issuer)
                        ')'
```
should work probably (?)

adelarja: @ebma, thanks for the comment.

I did it initially, but the issuer information is represented as bytes. I'll attempt to convert between bytes and SS58 addresses, as we've done in the past.

ebma: Oh... true I did not think about that. Let's not use SS58 here though because Stellar is not using that and it would spit out some info that is not useful in that context. Instead we should try to derive the Stellar public key from that raw bytes.

I think we should be able to use [this](https://github.com/stellar/js-stellar-base/blob/5ed52ab68d6c5305a10f4532087e08de75c814c5/src/address.js#L46) function to get the public key from the raw binary data. So we need to:
- add the `stellar-base` dependency (`npm install stellar-base`)
- import the `Address` type from `stellar-base` 
- use `Address.account(event.currencyId.value.issuer).toString()` to print the public key

I would suggest that we move this logic to derive the public key from the issuer to an extra function again. Something like
```
function deriveStellarPublicKeyFromRaw(raw: <type>) {
   const address = Address.account(event.currencyId.value.issuer);
   return address.toString()
}
```

src/mappings/token.ts
adelarja: Changed! Thanks for the comment @ebma

scripts/suscribeTransfer.js
adelarja: Added! Thanks for the comment

src/mappings/token.ts
adelarja: Done! :muscle:

ebma (approved):
Great job @adelarja 🎉  I added some minor changes but it's ready for merge now IMO

PR #25, opened 2023-11-01
Filter out the null bytes from asset code

### LINKED ISSUE #24: Fix error due to invalid encoding
The amplitude squid seems to get stuck because of some `CurrencyId`s that contain a null byte in their asset code.
image

We should filter this byte before putting the data into the database of the squid because apparently, it can't handle it.

Context for the error [here](https://stackoverflow.com/a/1348551).

ebma:
This PR adds a new function that filters out the null bytes from the asset code of a Stellar asset. It also slightly refactors the other function to derive the public key but the logic stays the same.

The amplitude squid is now able to sync properly.

Closes #24.


adelarja (approved):
Nice finding!

PR #29, opened 2023-11-08
Update manifests with new parameters

### LINKED ISSUE #22: Use dedicated deployments and reduce storage size of postgres database to cut costs
At the moment, the postgres storage is the most expensive factor in our billing and by a large margin.
We should try to cut down these costs by down-scaling the storage size of the database. This can be done by setting the `storage` parameter in the manifest, see [here](https://docs.subsquid.io/deploy-squid/pg-addon/#scaling). Note that the default parameter apparently is not 10 GB as written down in the documentation but 50 GB. 

Furthermore, we should configure our squids to be deployed on dedicated instances instead of 'collocated' development machines.

## TODO
Set the [storage](https://docs.subsquid.io/deploy-squid/pg-addon/#scaling) parameter to 10 GB and see if the squid still works as expected. Should be tested on Amplitude as it has the most blocks.
Set the [dedicated](https://docs.subsquid.io/deploy-squid/scale/#dedicated) parameter to true for all squids.

ebma:
Closes #22.

### DIFF
--- a/squid-amplitude.yaml
+++ b/squid-amplitude.yaml
@@ -1,6 +1,6 @@
 manifestVersion: subsquid.io/v0.1
 name: amplitude-squid
-version: 7
+version: 8
 description: 'Amplitude Kusama Squid'
 build:
 deploy:
@@ -31,7 +31,7 @@ scale:
     dedicated: true
     addons:
         postgres:
-            storage: 10G
+            storage: 1G
             profile: small
     processor:
         profile: small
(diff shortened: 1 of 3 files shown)


gianfra-t (approved):

ebma:
@gianfra-t can you push an update to this PR with the latest parameters we use in the manifests? I think we further reduced the storage parameter.

gianfra-t:
> @gianfra-t can you push an update to this PR with the latest parameters we use in the manifests? I think we further reduced the storage parameter.

Pushed. Yes, we reduced it to 1Gb and it is at around 10% usage.

ebma:
Thanks! Will merge it now

PR #32, opened 2023-11-20
Add remark to transfers

### LINKED ISSUE #30: Update the indexer to provide metadata/tags added with setRemark()
### Context
When subscribing to the indexer to monitor events like off ramp transaction, the stablecoin issuer has a need to track a reference number associated with transactions during the off-ramp process.

### Requirement

Update the indexer to also provide metadata/tags added with setRemark()  for executed transfers.

### Technical considerations

Unlike on Stellar, it's not possible to add a 'memo' to transfers by default. Nevertheless, we can achieve something similar by grouping calls to `tokens/balances::transfer()` and `system::remark()` extrinsic into the same transaction. Thus, we assume that users who want to attach reference numbers to their transfers group the calls to transfer and set a remark in a `utility::batch()` or `utility::batchAll()` extrinsic.

Thus, we need to change the indexer such that for every transfer that is [handled](https://github.com/pendulum-chain/pendulum-squids/blob/1cd1a27efdbe0f2ba45161f0ac39ecad0d6a2aa3/src/mappings/token.ts#L421), we also check if the same transaction contains a call to `system::remark()` and if so, include the contained data into the `Transfer` entity stored by the indexer.

gianfra-t:
Issue #30 

This PR will add the capability to the indexer of appending remarks that could work as "memos".

The `Transfer` and `TokenTransfer` entities now have a new `remark` field used for this purpose.

AFTER we handle events, we loop through all `batch` and `batchAll` calls and match the transfer a user has made with a corresponding remark on the same batch transaction, appending the remark data into the corresponding transfer remark field.

### Assumptions
- There exists only one transfer with same amount and destination on that block. Otherwise, only one (latest) would get the remark appended.
- The transfer events must always be processed and stored before we handle the batch calls. 
- Each `batch` and `batchAll` call must contain a remark and either a  `Balances.transfer` or a `Tokens.transfer` call. Any other combination of those kinds will be ignored, yet there can be more calls of different kind.

### Testing
This implementation was tested using a local chain both with `batch` and `batchAll`.


ebma (approved):
Great job @gianfra-t 👍  I tested this by locally running the squid and doing some batch transfers on Foucoco and the remark is appended for native and token transfers. 

Ready for merge IMO

ebma:
Considering your comment [here](https://www.notion.so/Automating-BRZ-EUR-on-Pendulum-4037d39f0b434a5ab2e299195b228b33?d=8688393247da45938b9a519a78786818&pvs=4#b781b071c6cf4ff889163d502adeaea5), we probably should not merge this just yet.

You raised the valid concern that if we have multiple transfers with multiple remarks in the same block _originating from the same sender_, we are not able to appropriately link remarks and transfers. I think the only way we can have this perfect linking is if we only link these calls if they are contained in a `utility::batch` or `utility::batchAll` extrinsic. As far as I know, these extrinsics also emit events, so it should be somewhat easy to move the linking to the handler of those events. What do you think @gianfra-t?

gianfra-t:
It would be ideal to group the two extrinsics using a `batch` event, but the event itself does not contain any information about what was executed. As far as a know and also according to [this post](https://substrate.stackexchange.com/questions/174/correlate-events-with-extrinsics-in-a-utility-batch-call-in-polkadot) the only way to do this would be to infer this by the order in which the batch events are executed.

But I am not sure if we can rely on the order 100% of the times when indexing. I will try to look on the subsquid documentation for this. In case we cannot rely on the order, an alternative I can think of would be to wait a block, and include the hash of the transfer into the remark,

ebma:
Hmm okay, thanks for pointing it out. So I take back that it should be easy to handle the batch calls. But it has to be possible to link all events and calls that happen in a batch extrinsic to exactly that extrinsic. Otherwise polkadot.js also wouldn't be able to do so, e.g. [here](https://polkadot.js.org/apps/?rpc=wss%3A%2F%2Fpencol-roa-00.pendulumchain.tech%2F#/explorer/query/0xe0b5b63ad72648065f59431a1bbeb82d406cf4c584bef8b039301eca281a4f5d).

As far as I understand we do have the information about what was executed in a batch transaction because we do know all the calls contained in it. 
image
We would have to find a way to group the [calls](https://github.com/pendulum-chain/pendulum-squids/pull/32/files#diff-7199f058b35aa392107dbd32106c45331e7f2fa9fc6de120e05e78d9db49f0faR278-R288) by their parent extrinsic, but I assume that this information is somehow contained on some attribute on each `call` entity.

gianfra-t:
I honestly don't know how Polkadot.js is grouping them. So to be sure the info is there in the event in squids I will try this  with a local chain using the utility pallet.

gianfra-t:
Okay so now I take back what I said haha. Although in the event itself there is no info, we have access to the call that triggered the event. And from the call we can extract both extrinsics!

src/mappings/remark.ts
```diff
@@ -3,49 +3,113 @@ import { CallHandlerContext } from '../processor'
...
+            )
+        } else {
+            return
+        }
+
+        let matchingBalanceTransfer = await ctx.store.findOne(Transfer, {
+            where: {
+                from: origin,
+                to: destAddress,
+                amount: balancesCall.value.amount,
```
ebma: Do we have more info that we can use to match it against a transfer? Maybe we also need to include something like the timestamp or block height? Just using origin, destination and amount to find a unique `Transfer` seems a bit too naive to me.

src/mappings/remark.ts
```diff
@@ -3,49 +3,113 @@ import { CallHandlerContext } from '../processor'
...
     } catch (e) {
         ctx.log.info(
             `Error when decoding remark '${remarkRaw}' with id ${ctx.call.id}.`
         )
     }
 
-    // Either we append the remark to balance transfer or token
-    if (matchingBalanceTransfer) {
+    // handle balances pallet transfer with remark
+    if (kinds.includes('Balances')) {
```
ebma, line 37: We have a lot of duplicated code in both if-else branches, let's try to deduplicate it with some refactoring.

src/mappings/remark.ts
```diff
@@ -3,49 +3,113 @@ import { CallHandlerContext } from '../processor'
...
-            `Error when finding the correct transfer for remark with id ${ctx.call.id}. Inconsistent number of transfers for user: ${origin}`
-        )
+    // filter out batch calls where there one transfer and
+    // one remark only
+    if (!hasSystemWithRemark(ctx.call)) {
         return
     }
+    const calls = ctx.call.args.calls
+    const kinds = calls.map((call: any) => call.__kind)
+    if (kinds.length != 2) return
```
ebma: We can consider lifting this restriction to `kinds.length < 2` so that we can also consider batch calls that for some reason contain more than a transfer and a remark. WDYT @gianfra-t?

src/mappings/remark.ts
```diff
@@ -3,49 +3,113 @@ import { CallHandlerContext } from '../processor'
...
+
+    const remarkCall = calls.find((call: any) => call.__kind === 'System')
```
ebma: Maybe we should explicitly check if it is a `remark` here, to rule out problems in edge cases where we have multiple different `System` extrinsics inside the batch.

src/mappings/remark.ts
gianfra-t: I incorrectly removed the block height. This should be looking for transactions only in the current block where the batch occurs. 
Assuming that, yes we could also filter for timestamp but I believe is going to be the same for all (in the block).

src/mappings/remark.ts
gianfra-t: It's possible but how do we deal with for instance 2 transactions and 2 remarks on the same batch? Or 2 transactions and 1 remark, etc.

src/mappings/remark.ts
gianfra-t: The call to `hasSystemWtihRemark` above should be checking that. But I will modify it so it checks if it is only one remark, then is true.

src/mappings/remark.ts
gianfra-t: Yes, I checked and the timestamp is the same for two tx. made in the same block.

src/mappings/remark.ts
ebma: Hmm right. Instead of limiting it to exactly two extrinsics, we could limit it to contain exactly one transfer and exactly one remark. But maybe that's overengineering for something that has no need yet anyway. So I'm fine if we keep this limit as is then.

gianfra-t: I think alternatively we could accept batch and batchAll with many calls, as long as there is one remark and one of the transfers. Let me know if that is okay.

ebma (commented):
Looks good 👍 can you bump the version in the manifest and for now deploy the foucoco one on aquarium? Once we see the foucoco one works we can consider deploying the rest as well.

src/mappings/remark.ts
ebma: Yes, I'd favor us doing it that way but I'm okay either way. Your call

ebma (approved):
Looks great 👍 I didn't test this locally but I'd appreciate if we can bump the version to `11` and deploy that squid to foucoco to see if it syncs.

gianfra-t:
It actually is deployed as 10, but I realized now the other PR @bogdanS98 is working on is also 10, so I will bump to 11 and wait for that one to merge.

ebma:
Ah okay, I was not sure which changes the v10 contains. 
I quickly tried testing it and I submitted [this](https://polkadot.js.org/apps/?rpc=wss%3A%2F%2Fpencol-roa-00.pendulumchain.tech%2F#/explorer/query/0xa1acf8a5d0fdc02d81d4f97d84808d3c030df11bb3f89bfa67cbfe1e03cbb34e) extrinsic on Foucoco but it doesn't seem like the indexer is picking it up, at least I don't see it for [this](https://squid.subsquid.io/foucoco-squid/v/v10/graphql) squid with this query
```
query MyQuery {
  transfers(where: {remark_isNull: false}) {
    remark
    id
  }
  tokenTransfers(where:{remark_isNull:false}) {
    remark
    id
  }
}
```

gianfra-t:
That is strange because locally it works. Let me investigate.

I made a [transaction](https://polkadot.js.org/apps/?rpc=wss%3A%2F%2Fpencol-roa-00.pendulumchain.tech%2F#/explorer/query/0x4f109b37145dceebe237fd22091a9e187d1773c113f473bc0b4279ad6db4c37c) and it appears to be working. 
Strange(er) it looks like your transfer itself is not even registered! (there is not transfer at block `746814`)

ebma:
> Strange(er) it looks like your transfer itself is not even registered! (there is not transfer at block 746814)

Ahh wow, that's a nice catch 👍 Then the problem was on my side because I just created a transfer from an account to itself which apparenty did not get executed or at least it didn't emit an event. (Which is smart because a transfer to itself is useless obviously but I did not expect it to get ignored). Nevermind then.

ebma (approved):
Great job 👍  ready for merge after https://github.com/pendulum-chain/pendulum-squids/pull/33 is merged.

ebma:
Oh one more thing, do you really want to check in the [build artifact](https://github.com/pendulum-chain/pendulum-squids/pull/32/files#diff-b66581cc44dd43986275dae807193e54c6b0e0ae41d243c8966bc121eedf83bb) or was it by accident? We did not do that so far

gianfra-t:
> Oh one more thing, do you really want to check in the [build artifact](https://github.com/pendulum-chain/pendulum-squids/pull/32/files#diff-b66581cc44dd43986275dae807193e54c6b0e0ae41d243c8966bc121eedf83bb) or was it by accident? We did not do that so far

Accident... sorry. I will remove it before merging!

ebma:
Alright 👍  since the other PR is merged now we need to resolve the conflicts again

gianfra-t:
I've deployed v11 so once it syncs I think this is ready.

ebma (approved):

gianfra-t:
I [commented out](https://github.com/pendulum-chain/pendulum-squids/pull/32/commits/9af597e2ae559ebfddcdd71e55bf67f905d2a14c) the fetching of the max block height because I was getting an error `ReferenceError: fetch is not defined at...` when deploying the squids. 
This does not appear when running locally, perhaps `fetch` is not available on the squids environment. 

Let me know what you think @bogdanS98. Did you also get this error?. I think it could still work just by saving the blocks once we reach the latest block. Only downside is that as soon as it starts, the latest blocks will not be available.

bogdanS98:
@gianfra-t @ebma I only tested locally after adding `fetch()` and it worked. Global `fetch()` only works on Node.js versions starting from **18**. I use 18.8.2 locally but in the squid environment we have version **16**. There are a few options here:
1. We upgrade Node.js version in the squid environment and then we'll have global `fetch()` available
2. We add **node-fetch** module and use `fetch()` from there
3. We use something else instead of fetch but that still requires a module to be imported (e.g. axios)

I would go with the second option but please let me know what you think

gianfra-t:
I also would go with the second option since fetch already works and upgrading node to 18 may break other things. I will try this later.

gianfra-t:
We have replaced from using fetch to `axios` which seems to work on the cloud environment. I will wait for full sync to merge this PR though.

bogdanS98 (approved):

PR #33, opened 2023-11-21
Add info about basic general blockchain components to indexer

### LINKED ISSUE #31: Add info about basic general blockchain components to indexer
Our indexer currently does not offer much 'general' data. You cannot query the last blocks, extrinsics or events. The base version that is deployed in combination with the archive already exposes this data, see e.g. [here](https://foucoco.explorer.subsquid.io/graphql). The code for that 'base squid' is [here](https://github.com/subsquid-labs/giant-squid-explorer).
Sometimes we do need to have access to this basic data. For example, solving https://github.com/pendulum-chain/tasks/issues/60 requires it.

## TODO
We should extend the schema of our squids to also offer data about:
- blocks
- extrinsics
- calls
- events

We should **not** add the `metadata` as it takes too much space and we will never query it from these indexers. 

The implementation should draw heavy inspiration from the implementation of the [giant-squid](https://github.com/subsquid-labs/giant-squid-explorer). We should:
- adopt the necessary entities from the [schema](https://github.com/subsquid-labs/giant-squid-explorer/blob/main/schema.graphql)
- regenerate the types and models
- adopt the fields that are defined in [processor.ts](https://github.com/subsquid-labs/giant-squid-explorer/blob/main/src/processor.ts) and add the processing defined [here](https://github.com/subsquid-labs/giant-squid-explorer/blob/a3d9b4f08ecbd918f18c87fc715e742cbdf145c4/src/main.ts#L6) to our processor as well

bogdanS98:
Closes #31


src/processor.ts
```diff
@@ -92,170 +118,191 @@ const processor = new SubstrateBatchProcessor()
...
                         await handleUpdatedPrices({
                             ...ctx,
-                            block: block.header,
-                            event: item,
+                            block: block,
+                            event: event,
                         })
                         break
                     default:
+                        await saveEvent(ctx, event)
```
ebma: Wouldn't we want to save the event in any case, even if it was handled in one of the other cases? If so, we should move the call to save it out of the switch-case.

src/processor.ts
```diff
@@ -35,27 +37,51 @@ import {
...
             tip: true,
-        },
-        block: {
-            timestamp: true,
+            version: true,
         },
     })
+    .addCall({
+        extrinsic: true,
+        stack: true,
```
gianfra-t, line 83: Just curious, any idea what `stack` in the selector is?

gianfra-t (approved):
Seems all good to me. I tested in locally indexing foucoco and was able to see blocks, events, calls, etc.
Any idea why some blocks will return more calls than extrinsic count? I thought each call was associated with one extrinsic.

ebma:
> Any idea why some blocks will return more calls than extrinsic count? I thought each call was associated with one extrinsic.

I guess batching is a good explanation to this. If you batch two extrinsics (or rather 'calls') into the `utility.batch` or `utility.batchAll` extrinsic, then you have two calls but one extrinsic. For example, here I batched two calls to `system.remark` but as you can see on the side, it's contained in just one extrinsic.
image

src/processor.ts
ebma: In the [documentation](https://docs.subsquid.io/substrate-indexing/setup/data-requests/#calls) it says that 

> with stack = true it will add all calls in the stack of each matching call, including itself, to the calls iterable;

but I honestly don't really understand what that is supposed to mean as it's too ambiguous.

ebma: The 'base template' is [using it like this](https://github.com/subsquid-labs/giant-squid-explorer/blob/a3d9b4f08ecbd918f18c87fc715e742cbdf145c4/src/processor.ts#L60-L62) so it should be safe if we do too. But I can't tell the actual implications

gianfra-t:
> > Any idea why some blocks will return more calls than extrinsic count? I thought each call was associated with one extrinsic.
> 
> I guess batching is a good explanation to this. If you batch two extrinsics (or rather 'calls') into the `utility.batch` or `utility.batchAll` extrinsic, then you have two calls but one extrinsic. For example, here I batched two calls to `system.remark` but as you can see on the side, it's contained in just one extrinsic. image

Ohh I see now, thanks for the explanation.

src/processor.ts
```diff
@@ -133,27 +133,56 @@ export interface EventHandlerContext extends Ctx {
...
+    .then((response) => response.json())
+    .then((data) => {
+        console.log('Max height:', data)
+        maxHeight = data
+    })
+    .catch((error) =>
+        console.error('Error getting block count from archive:', error)
+    )
+
+const blockRetention = process.env.BLOCK_RETENTION
```
ebma: We now define this value twice in two different files. Let's rather move it to `config.ts` or something similar.

src/processor.ts
```diff
@@ -133,27 +133,56 @@ export interface EventHandlerContext extends Ctx {
     event: Event<Fields>
 }
 
+// In case call to archive fails, this should make block.height >= maxHeight - blockRetention false
+let maxHeight = Number.MAX_SAFE_INTEGER
+
+// Fetch max height from the archive
+fetch(config.dataSource.archive + '/height')
```
ebma: Let's also put this somewhere else.

.env
```diff
@@ -2,3 +2,4 @@ DB_NAME=squid
 DB_PORT=23798
 PROCESSOR_PROMETHEUS_PORT=3000
 GQL_PORT=4350
+BLOCK_RETENTION=7200
```
ebma: Maybe we should add a word to the name that this indicates being a number.
`BLOCK_RETENTION_NUMBER` or `LAST_BLOCKS_RETENTION_NUMBER`

gianfra-t (approved):
Seems good !

src/processor.ts
```diff
@@ -92,170 +118,212 @@ const processor = new SubstrateBatchProcessor()
...
+
+    // Fetch max height from the archive
+    const maxHeight = await maxHeightPromise
+
+    for (let { header: block, calls, events, extrinsics } of ctx.blocks) {
+        ctx.log.debug(
+            `block ${block.height}: extrinsics - ${extrinsics.length}, calls - ${calls.length}, events - ${events.length}`
+        )
+
+        if (isHead || block.height >= maxHeight - blockRetention) {
```
ebma: Let's add a descriptive comment above this statement so we don't forget what we are doing here in a week or two.

ebma (approved):
Looks good 👍  ready for merge, I'd just like to have one more code comment

src/processor.ts
gianfra-t: Perhaps `isHead` is not even needed, since when that is true, the statement on the left will also be right?

bogdanS98: You're right, `isHead` is not needed here. `block.height >= maxHeight - blockRetention` is always going to be true for the last `blockRetention` blocks synced and after syncing is done

ebma (approved):

PR #37, opened 2023-12-15
Add condition to only process price updates for the most recent block

### LINKED ISSUE #34: Only process price info updates for the latest blocks
At the moment, the syncing of our squids is significantly slowed down by processing price update events. Since we don't care about historic prices but on every block, we just overwrite the latest price, we can speed this up by only processing these events if the current block is the 'head' of our chain.

bogdanS98:
Closes #34.


src/processor.ts
```diff
@@ -145,6 +145,11 @@ processor.run(new TypeormDatabase(), async (ctx) => {
     // Fetch max block height from the archive
     let maxHeight = await maxHeightPromise
 
+    // Determine the tallest block in the batch
+    let tallestBlockHeight = Math.max(
```
ebma: 'tallest' sounds weird haha. Can we rename it to `highestBlockHeight` or similar? Tall sounds so human.
Maybe it's actually related to check if `ctx.isHead` in the condition below and not checking the block height. Should already improve it a lot, WDYT @bogdanS98?

bogdanS98: Indeed it would already improve by a lot, but `ctx.isHead` will always be true after the endpoint switches from archive to RPC (real-time) because every new batch that comes through will contain the head. So if we don't check that the block we're currently processing is indeed the head (has the biggest height) then we would always save the prices for every block in the batch.  We could run it with and without the second condition to see if the difference is significant, how does that sound to you @ebma?

src/processor.ts
```diff
@@ -145,6 +145,11 @@ processor.run(new TypeormDatabase(), async (ctx) => {
     // Fetch max block height from the archive
     let maxHeight = await maxHeightPromise
 
+    // Determine the tallest block in the batch
+    let latestBlockHeight = Math.max(
```
bogdanS98: Is this clear enough? WDYT @ebma?

src/processor.ts
ebma: But I assume that once we are connected to the RPC (ie are getting blocks in real-time) then the batch would always just contain 1 block (?) And if that's the case, we don't need to check after all. 

Can you please check the batch sizes that we get once connected to the RPC? Then we base our decision on that info.

bogdanS98: Logged the batch size and you're right, after switching to RPC the batch always contains only the latest block so we can remove the additional condition
image

bogdanS98: Updated the code 👍🏼

src/processor.ts
bogdanS98: This is not needed anymore

ebma (approved):
Nice 👍 please deploy them to aquarium and once we confirmed that they work as expected bump them to be the production squids.

bogdanS98:
Deployment was successful and catch up time periods decreased from 1h+ to an average of 10 minutes 👍🏼

PR #38, opened 2023-12-15
Composition zenlink pair not shown properly

### LINKED ISSUE #36: Composition of Zenlink pair not shown properly
The Zenlink UI currently shows a pool composition of 2014450 AMPE to 0 KSM for the [AMPE/KSM pool](https://app.zenlink.pro/pool/amplitude:2124-2-1099511628544). This is not true as the `pairAccount` of the pool `6jM63XCYUjHhdfXy9YMa2fKa9z4nqyJTb4DQxjm5mqsXBBQ5` has a balance of ~100 KSM. 

image

image

## TODO
We need to investigate this issue and make sure that the `reserve1` parameter of the following query is not `0` but displays the amount of KSM that the pool has.

```
query MyQuery {
  pairById(id:"2124-2-1099511628544") {
    id
    reserve0
    reserve1
    reserveETH
    reserveUSD
    token0Price
    token1Price
    totalSupply
    trackedReserveETH
    volumeToken0
    volumeToken1
    volumeUSD
    liquidityProviderCount
    txCount
    untrackedVolumeUSD
  }
  bundleById(id: "1") {
    id
    ethPrice
  }
}

```

gianfra-t:
closes #36 

Problem was that new types `v12` for storage (and events) where implemented in amplitude given a runtime upgrade. 
This caused some events to be ignored so that the indexer was not giving the correct values for liquidity

Fococo was also changed to `v4`, so events and storage types were also changed.

Some minor [modification](https://github.com/pendulum-chain/pendulum-squids/pull/38/files#diff-1e1d77462c947d5a2ed268da655215c899db2cfb25a6cbaac56dc3e2ea776a37R19) was made to support the new CurrencyId (with token variant) that comes from `v12` in amplitude.


ebma (approved):
Thanks for these changes, looks good overall 👍 
We could slightly improve the code by using if-elseif branches instead of multiple `if`s because the code could exit earlier instead of doing multiple checks. But this is super minor and probably not worth refactoring right now. 

Having a function like the `decodeFoucocoEvent` would be nice because I agree that all the version checks can get super annoying in the future. But I agree that this should be done in a separate PR. Could you create a respective issue ticket for this @gianfra-t?

gianfra-t:
> We could slightly improve the code by using if-elseif branches instead of multiple ifs because the code could exit earlier instead of doing multiple checks. But this is super minor and probably not worth refactoring right now.

True! let's do this in the new refactor issue, I don't think we will need this anyway since the idea is to get rid of the `if`s.
I will merge this now.

ebma (approved):

PR #40, opened 2023-12-18
refactor all calls to storage and event decoding

### LINKED ISSUE #39: Refactor event and storage version selection
Right now, when decoding and event or querying the storage, we are checking for the corresponding storage and event version one by one like [here](https://github.com/pendulum-chain/pendulum-squids/blob/main/src/utils/token.ts#L375).

Given that in the near future we will most likely introduce several new versions, we should refactor the code and add some function like the one introduced in [this PR](https://github.com/pendulum-chain/pendulum-squids/pull/38/files#diff-15b7fe769e3c2639a3fb1370b9d007ad4ed46d36bf68a6834776fe7d2bf11d5dR6) that allows to select automatically the storage/event version corresponding to the context, and decode it.

gianfra-t:
Issue #39

Besides the changes, there is also a bug fix for the foucoco squid.


src/types.ts
```diff
@@ -21,3 +34,59 @@ export interface ProcessorConfig {
...
+            moduleName,
+            storageEntity,
+            ctx
+        )
+    }
+
+    if (versionedStorage) {
+        return versionedStorage
+    } else {
+        throw new Error('Storage for could not be found')
```
ebma:
```suggestion
        throw new Error(`Storage for ${network} could not be found`)
```

ebma (approved):
Nice! Some really great changes, I love it 🙏

ebma (approved):

gianfra-t:
I will test a few queries on the deployed v13 squids and then merge this.

PR #42, opened 2024-01-24
Research how reliable past diadata is

ebma:
Related to https://github.com/pendulum-chain/tasks/issues/192. 

Please refer to the [notion](https://www.notion.so/satoshipay/DIA-price-accuracy-analysis-b3c9ea6f325e4e599198bf3717a3d282) page for detailed explanation and discussion. 

Not intended to be merged, ever. This PR is just for illustrative purposes so that the research results can be retraced.

## How to run

```
npm install
sqd build
sqd down # make sure database is empty so we process from scratch
sqd up
sqd process:foucoco # analyse the prices on Foucoco
```

Once the squid has finished fetching all price events, and every X blocks, a new `.json` file (defined here) will be generated in the root directory.

NOTE: if the squid is stopped, the deviations result file will not persist. The squid must sync from beginning to the highest block.

Check the [notion](https://www.notion.so/satoshipay/DIA-price-accuracy-analysis-b3c9ea6f325e4e599198bf3717a3d282) page, "Results on Amplitude" for the latest file with the collected data to avoid re-running the squid.

Run `deviation_analysis.py` for final plot and analytics. Installing python with [Anaconda](https://www.anaconda.com/) already has the required dependencies (Pandas, Matplotlib).

## What's analyzed

The processing of the squid was changed. The idea is the following:
- the squid will start processing the blocks from block `0`. The data is from one of our Foucoco archive nodes.
- each block, the squid checks for price feeding events emitted by the DIA oracle pallet
- if there is a price event, the squid starts the analysis of the coin info supplied to the chain, saving it to a file for post-analysis and plotting.

The price data that is provided to chain from the offchain worker is exposed by the batching server. And the batching server receives the data from one of the many [data sources.](https://www.diadata.org/app/source/defi/)

## Results

Find context and results in [this](https://www.notion.so/satoshipay/DIA-price-accuracy-analysis-b3c9ea6f325e4e599198bf3717a3d282) Notion page.

### DIFF
--- a/deviation_analysis.py
+++ b/deviation_analysis.py
@@ -51,7 +51,7 @@ for currency, records in data.items():
     print(f"{currency} - Analysis to: {last_timestamp}")
     print(f"{currency} - Deviation events: {count}")
     print(f"{currency} - Total blocks: {all_records_count}")
-    print(f"{currency} - Accuracy: {accuracy}%")
+    print(f"{currency} - Accuracy: {accuracy * 100}%")
     print(f"{currency} - Mean Deviation: {mean_deviation:.2f}%")
     print(f"{currency} - Mean Deviation of Deviations: {mean_of_deviation:.2f}%")
     print(f"{currency} - Standard Deviation of Deviations: {std_deviation_of_deviation:.2f}%")
--- a/deviation_analysis.py
+++ b/deviation_analysis.py
@@ -1,26 +1,22 @@
+# script to run to analyze the deviation data. Histograms, boxplots, and metrics are generated for each currency analyzed.
 import pandas as pd
 import matplotlib.pyplot as plt
 import json
 
-with open('dev.json', 'r') as file:
+with open('dev_amplitude.json', 'r') as file:
     data = json.load(file)
 
 dfs = {currency: pd.DataFrame(records) for currency, records in data.items()}
 
 for currency, df in dfs.items():
-    # Original count of rows
     original_count = df.shape[0]
     
-    # Remove rows where deviation is null
     df_cleaned = df.dropna(subset=['deviation'])
     
-    # Count of rows after removal
     cleaned_count = df_cleaned.shape[0]
     
-    # Calculate how many rows were removed
     rows_removed = original_count - cleaned_count
     
-    # Proceed with the cleaned DataFrame
     # Histogram
     plt.figure(figsize=(10, 6))
     df_cleaned['deviation'].plot(kind='hist', bins=100, title=f'Deviations Histogram for {currency}')
@@ -28,7 +24,7 @@ for currency, df in dfs.items():
     plt.ylabel('Frequency')
     plt.show()
 
-    # Boxplot
+    # Box plot
     plt.figure(figsize=(10, 6))
     df_cleaned.boxplot(column=['deviation'])
     plt.title(f'Boxplot of Deviations for {currency}')
@@ -37,11 +33,14 @@ for currency, df in dfs.items():
 
     # Metrics
     count = df_cleaned['deviation'].count()
+    first_timestamp = df_cleaned['timestamp'].min()
+    last_timestamp = df_cleaned['timestamp'].min()
     mean_deviation = df_cleaned['deviation'].mean()
     std_deviation = df_cleaned['deviation'].std()
-    
-    # Print metrics and count of null deviation rows removed
-    print(f"{currency} - Number of Deviations: {count}")
+
+    print(f"{currency} - Analysis from: {first_timestamp}")
+    print(f"{currency} - Analysis to: {last_timestamp}")
+    print(f"{currency} - Deviation events: {count}")
     print(f"{currency} - Average Deviation: {mean_deviation:.2f}%")
     print(f"{currency} - Standard Deviation of Deviations: {std_deviation:.2f}%")
     print(f"{currency} - Null deviation rows removed (lack of data): {rows_removed}\n")
--- a/deviation_analysis.py
+++ b/deviation_analysis.py
@@ -15,7 +15,7 @@ for currency, df in dfs.items():
     
     # Filter out rows where deviation is NaN or less than the threshold
     df_cleaned = df.dropna(subset=['deviation'])
-    df_cleaned = df_cleaned[df_cleaned['deviation'] > THRESHOLD]
+    #df_cleaned = df_cleaned[df_cleaned['deviation'] > THRESHOLD]
     
     cleaned_count = df_cleaned.shape[0]
     
--- a/deviation_analysis.py
+++ b/deviation_analysis.py
@@ -1,0 +1,47 @@
+import pandas as pd
+import matplotlib.pyplot as plt
+import json
+
+with open('dev.json', 'r') as file:
+    data = json.load(file)
+
+dfs = {currency: pd.DataFrame(records) for currency, records in data.items()}
+
+for currency, df in dfs.items():
+    # Original count of rows
+    original_count = df.shape[0]
+    
+    # Remove rows where deviation is null
+    df_cleaned = df.dropna(subset=['deviation'])
+    
+    # Count of rows after removal
+    cleaned_count = df_cleaned.shape[0]
+    
+    # Calculate how many rows were removed
+    rows_removed = original_count - cleaned_count
+    
+    # Proceed with the cleaned DataFrame
+    # Histogram
+    plt.figure(figsize=(10, 6))
+    df_cleaned['deviation'].plot(kind='hist', bins=100, title=f'Deviations Histogram for {currency}')
+    plt.xlabel('Deviation (%)')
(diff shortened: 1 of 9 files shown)


ebma:
@pendulum-chain/devs please have a look at the notion page linked in the description. It contains first results and some evaluation of them. Depending on your feedback, I can change/extend the logic for the analysis if needed.

TorstenStueber:
@ebma I left a few comments on the notion page.

I did not execute this branch locally. Can you quickly explain how it works when you run it? Does it ingest all blocks from the beginning and where is the data from? From the RPC nodes or are all the blocks mirrored on some Subsquid server and it queries them from there?

ebma:
Thanks for the comments. I updated the description.

gianfra-t:
@TorstenStueber @ebma I ran the the analysis again with a slightly different approach ([version 2](https://www.notion.so/satoshipay/DIA-price-accuracy-analysis-b3c9ea6f325e4e599198bf3717a3d282?pvs=4#61763950e0d74603a72582949f92960a) in the notion), the results are also there. There may be a mistake with the analysis or the assumptions, but I see some some very large deviations with the Kraken data particularly for KSM and XLM, which makes some sense given the higher volatility compared to the FIAT assets, but still should not be that high.

I could not get data for the remaining FIAT currencies.

src/mappings/prices.ts
```diff
@@ -16,37 +17,64 @@ const KrakenKsmUsd = readCSV(
...
+            const price_deviation =
+                ((ohlcvAtTime!.low - diaPrice) / ohlcvAtTime!.low) * 100
+            let deviation: PriceDeviation = {
+                timestamp,
+                deviation: price_deviation,
+            }
+            saveDeviation(deviation, symbol)
+        } else if (diaPrice > ohlcvAtTime!.high) {
+            const price_deviation =
+                ((diaPrice - ohlcvAtTime!.high) / ohlcvAtTime!.high) * 100
```
ebma, line 83: I think it's more accurate to divide `(a - b) / a` like we did in l.63 
```suggestion
                ((diaPrice - ohlcvAtTime!.high) / diaPrice) * 100
```

deviation_analysis.py
```diff
@@ -0,0 +1,52 @@
...
+    data = json.load(file)
+
+dfs = {currency: pd.DataFrame(records) for currency, records in data.items()}
+
+for currency, df in dfs.items():
+    original_count = df.shape[0]
+    
+    # Filter out rows where deviation is NaN or less than the threshold
+    df_cleaned = df.dropna(subset=['deviation'])
+    #df_cleaned = df_cleaned[df_cleaned['deviation'] > THRESHOLD]
```
ebma: So for your analysis results v2 in the Notion page, we are not filtering out any values based on this `THRESHOLD`, correct?

deviation_analysis.py
```diff
@@ -0,0 +1,52 @@
...
+import json
+
+THRESHOLD=0.1
+
+with open('dev_amplitude.json', 'r') as file:
+    data = json.load(file)
+
+dfs = {currency: pd.DataFrame(records) for currency, records in data.items()}
+
+for currency, df in dfs.items():
```
ebma: As far as I understand, this dataframe is only filled if the price is not in the range 'ohlcvLow' < price < 'ohlcvHigh', see the conditions [here](https://github.com/pendulum-chain/pendulum-squids/pull/42/files#diff-08d6fa9cad6cb45bbcb2efc2f6f79e5c635d49266aa9b3087b27f8d1884b5506R61-R69). This makes sense to keep the dataframe low in size. 

However, this means that the 'mean deviation' that is calculated below is the 'mean deviation **of deviations**' and not of all prices since all the datapoints that are between 'low' and 'high' are not contained in the dataframe and not taken into account when calculating the average. I think that this is the reason why the results for the average deviations in your analysis are so high. It would be interesting to see how large the average deviation is when taking all prices into account not just the outliers.

deviation_analysis.py
gianfra-t: Correct, no filtering since it can be done in the python script and avoid re-running.

src/mappings/prices.ts
gianfra-t: You are right! Missed that.

deviation_analysis.py
gianfra-t: Exactly, It means the mean deviation of deviations just like the standard deviation is also of deviations, I forgot to mention that in the print line. It is not a measure of the overall accuracy itself. It tells us that if the price deviates, it will deviate by that much, but not necessarily a measure of how many times it deviates. Sadly the deviation events count is already very high in comparison with all the blocks analyzed.

To get the average deviation accounting for all blocks, we can in the python analysis script add all the entries for the remaining blocks assuming they deviation is 0 and recalculate the mean, which is a conservative but reasonable estimate given that the price was in between the high,low boundary.

I will modify so we get an account of how many blocks exactly were analyzed. This is possible to do right now from the difference in timestamp and assuming 12 seconds per block, but it may be better to have the number explicit.

src/mappings/prices.ts
gianfra-t: I now think I prefer the other way around, because I think we may want the percentage above (or below) the "true" value or limit, which is `ohlcvAtTime!.high`. What do you think @ebma ?

In the other case it is inverted because the value is more negative.

ebma: Ah hmm, well you have a point. I'm fine if we keep it like this then 👍

deviation_analysis.py
```diff
@@ -0,0 +1,86 @@
...
+    accuracy = ((all_records_count - count) / all_records_count) 
+    mean_deviation = all_records['deviation'].mean()
+    mean_of_deviation = df_cleaned['deviation'].mean()
+    std_deviation_of_deviation = df_cleaned['deviation'].std()
+
+    print(f"{currency} - Analysis from: {first_timestamp}")
+    print(f"{currency} - Analysis to: {last_timestamp}")
+    print(f"{currency} - Deviation events: {count}")
+    print(f"{currency} - Total blocks: {all_records_count}")
+    print(f"{currency} - Accuracy: {accuracy}%")
```
TorstenStueber: We should probably multiply by 100 here.

fiat_pair_former.py
```diff
@@ -0,0 +1,25 @@
...
+csv_file_path_2 = 'ohlcv/ETHUSD_1.csv'
+
+result_csv_file_path = 'EURUSD.csv'
+
+df1 = pd.read_csv(csv_file_path_1, header=None)
+df2 = pd.read_csv(csv_file_path_2, header=None)
+
+result_df = df1.iloc[:, [0]]
+
+divided_columns = df2.iloc[:, 1:] / df1.iloc[:, 1:]
```
TorstenStueber, line 21: To be totally correct, the low and high prices would result from cross division instead of pointwise division (e.g., low EURUSD = low ETHUSD / high ETHEUR instead of low ETHUSD / low ETHEUR) but at the end the results would look even better than they already are – so we won't win a lot here.

TorstenStueber:
> There may be a mistake with the analysis or the assumptions, but I see some some very large deviations with the Kraken data particularly for KSM and XLM, which makes some sense given the higher volatility compared to the FIAT assets, but still should not be that high.
> 
@gianfra-t I actually find the XLM and EUR-USD derivation you found quite okay. The AUS-USD on the other hand is quite bad. Do we know more about this? Is it fiat AUS-USD or is it some kind of stable coin prices traded within Kraken? There is a good chance that Kraken is off here instead of DIA.

gianfra-t:
I assume it is FIAT [prices](https://pro.kraken.com/app/trade/aud-usd), since token prices would specify the name or contract of the token. Since `dia` prices are a collection of scrappers/sources maybe we should relax our assumption that Kraken data is the ground truth and treat it more as a sanity check.

fiat_pair_former.py
gianfra-t: You are right! But keep in mind that this code was not used for analysis, since either the FIAT pairs were already on the data, or there were no FIAT/Crypto pairs to form the missing ones. I left the code just as a reminder in case we needed to form a FIAT/FIAT pair someday. 

Do you think it makes sense to keep it around or just delete it since it may create confusion?

TorstenStueber: Ah okay, for me it was not clear how your code was used. It's probably good to either add some documentation to the code in this branch or document here in this non-mergable PR how those files have been used. I could figure that you first run the squid from scratch which then created a JSON file, which you then analyzed with one of the python scripts. Or is there this kind of documentation and I missed it?

gianfra-t: Yes the process is like that. The notion and the description of this PR should already explain the process. I will try to make it more clear.

ebma:
Closing this as our research is done and this was never intended to get merged.

PR #49, opened 2024-02-02
refactor storage and event selector

### LINKED ISSUE #43: Refactor storage and event decoders
After the addition of #40 , a bug was introduced where regenerating the types would delete the the new decoder [files](https://github.com/pendulum-chain/pendulum-squids/blob/main/src/types/pendulum/eventsAndStorageSelector.ts). 

We need to move this files out of the `types` folder, and replace the corresponding imports.

Also, it  would be a nice addition to just define one decoder function that works for every chain, perhaps by passing the network identifier.

gianfra-t:
This PR is related to the bug described #43 here, and also removes duplicate definition of the event decoder and storage functions.


gianfra-t:
@pendulum-chain/devs let me know what you think, and then I can try and deploy the squids, and compare some queries.

src/eventsAndStorageSelector.ts
```diff
@@ -0,0 +1,90 @@
...
+const pendulumStorageVersions: string[] = ['v1', 'v3', 'v9', 'v10']
+const amplitudeEventVersions: string[] = ['v1', 'v3', 'v7', 'v8', 'v10', 'v12']
+const amplitudeStorageVersions: string[] = [
+    'v1',
+    'v3',
+    'v7',
+    'v8',
+    'v10',
+    'v12',
+]
```
ebma: I think we can always assume that event and storage versions will be exactly the same for each network. So we can reduce this to one array per network.

src/eventsAndStorageSelector.ts
```diff
@@ -0,0 +1,90 @@
...
+    if (network === 'foucoco') {
+        networkEventsAny = foucocoEvents as { [key: string]: any }
+        eventVersions = foucocoEventVersions
+    } else if (network === 'pendulum') {
+        networkEventsAny = pendulumEvents as { [key: string]: any }
+        eventVersions = pendulumEventVersions
+    } else {
+        networkEventsAny = amplitudeEvents as { [key: string]: any }
+        eventVersions = amplitudeEventVersions
+    }
```
ebma: Instead of returning amplitude by default, we could add another `else if` branch for amplitude and throw an error in the return. This is super minor and should never happen but if for example one day we introduce a typo in one of the `network` env variables, then this check would catch that error instead of silently decoding the events for amplitude.

src/eventsAndStorageSelector.ts
```diff
@@ -0,0 +1,90 @@
...
+): Promise<StorageTypeFoucoco | StorageTypePendulum | StorageTypeAmplitude> {
+    let networkStorageAny
+    let storageVersions: string[]
+    if (network === 'foucoco') {
+        networkStorageAny = foucocoStorage as { [key: string]: any }
+        storageVersions = foucocoStorageVersions
+    } else if (network === 'pendulum') {
+        networkStorageAny = pendulumStorage as { [key: string]: any }
+        storageVersions = pendulumStorageVersions
+    } else {
```
ebma: Same here

src/eventsAndStorageSelector.ts
ebma, line 1: Maybe it's better to move this file to `src/types` instead.

ebma (commented):
Great changes 👍  I left some minor comments how we might further improve it but it should be good to go overall

src/eventsAndStorageSelector.ts
gianfra-t: mmm I didn't know about that. I will reduce it.

src/eventsAndStorageSelector.ts
gianfra-t: Okay, I thought this was also deleted when generated. Let me test.

ebma (approved):
I didn't test it locally but the changes look good to me 👍

gianfra-t:
I saw some errors in foucoco only, from contract events. But I think those were before already.

ebma:
Depends on the error you saw of course but I think @bogdanS98 also encountered contract errors on Foucoco so it's probably nothing that got introduced in this PR.

gianfra-t:
Yes I also remember exploring these errors in Foucoco some months ago. Most likely they are the same errors. Since the sync is now complete, I will point to these new squids and stop the old ones.

ebma:
Alright 👍  ready for merge IMO

PR #51, opened 2024-02-08
50 fix constants to enable indexing of new pairs and pools

### LINKED ISSUE #50: Fix constants to enable indexing of new pairs and pools
After the [addition](https://github.com/pendulum-chain/transaction-builder/pull/57) of new swap pairs for Amplitude and Pendulum, we need to make some changes to the hard-coded constants in the processor required to properly index the events.

For instance, [here](https://github.com/pendulum-chain/pendulum-squids/blob/2c892a4cf063ef105dc85ca43b6f91e3ae60d354/src/constants.ts#L19) we need to add all the relevant tokens that are used in the pairs on both chains.

We also need to ensure that relevant types (farming, zenlink) are generated for Pendulum.

Finally, we need to modify the use of the hard-coded `CHAIN_ID` [constant](https://github.com/pendulum-chain/pendulum-squids/blob/2c892a4cf063ef105dc85ca43b6f91e3ae60d354/src/constants.ts#L58) currently pointing to Amplitude.

gianfra-t:
Closes #50.

 ### Changes in this PR
- [Addition and modification](https://github.com/pendulum-chain/pendulum-squids/pull/51/files#diff-8fa4b52909f895e8cda060d2035234e0a42ca2c7d3f8f8de1b35a056537bf199L31) of token addresses in `TOKEN_METADATA_MAP`, which is [required](https://github.com/pendulum-chain/pendulum-squids/blob/2c892a4cf063ef105dc85ca43b6f91e3ae60d354/src/entities/token.ts#L21) to properly handle any incoming token, otherwise the event with will be ignored or wrongly indexed.
- [Addition](https://github.com/pendulum-chain/pendulum-squids/pull/51/files#diff-8fa4b52909f895e8cda060d2035234e0a42ca2c7d3f8f8de1b35a056537bf199R88) of Pendulum's `CHAIN_ID`, and corresponding getter function.
- [Modification](https://github.com/pendulum-chain/pendulum-squids/pull/51/files#diff-21b6ee42f512ec2382d4507d41315108e36137b2a99455f6328d9c004e7767c2R202) of the `handleTokenDeposited` function. Previously, for any deposit related to `ZenlinkLPToken` we were creating a `Burn` addition to the `transaction` entity, when it should be a `Mint` that represents the creation of said LP token. Please, see [bifrost implementation](https://github.com/zenlinkpro/bifrost-kusama-squid/blob/main/src/mappings/token.ts#L111) for the same handler.
This was leading to problems when indexing a liquidity addition event, where a `mint` entry in the transaction is [expected](https://github.com/pendulum-chain/pendulum-squids/blob/2c892a4cf063ef105dc85ca43b6f91e3ae60d354/src/mappings/protocol.ts#L162). 
- Addition of farming types to `typegen-pendulum.json` and regeneration of types.
- [Addition](https://github.com/pendulum-chain/pendulum-squids/pull/51/files#diff-65d795c879e5c4eb1484725cb05497311596d0aad76811d2e7bb9c1bbd99e4f6R13) of new tokens to WHITELIST.

### Other changes
- In function `killFarmingPoolInfo`, we [modify](https://github.com/pendulum-chain/pendulum-squids/pull/51/files#diff-4d4bcc1a88df8dc5cc1ddaade11bb616d118b89bca2bd4e9e3fe353a76c13712R453) the condition to handle `ZenlinkLPToken` token kind instead of `LPToken`, and transform it to zenlink id in the same way as `updateFarmingPoolInfo`.
This change is not necessary for the issue and sadly, there are no "kill pool" events to test it. 
- Also in lines with the previous change, we [remove](https://github.com/pendulum-chain/pendulum-squids/pull/51/files#diff-1e1d77462c947d5a2ed268da655215c899db2cfb25a6cbaac56dc3e2ea776a37L42) unnecessary enums and objects.


src/utils/pricing.ts
```diff
@@ -13,6 +13,15 @@ export const KSM = '2124-2-256'
 export const WHITELIST: string[] = [
     '2124-0-0', // wnative
```
ebma, line 42: Should we also whitelist the native token on Pendulum ie `2094-0-0`? The check in `findEthPerToken()` should then probably check whether it's the native token on Pendulum or Amplitude and then return `ONE_BD`.

src/utils/token.ts
```diff
@@ -39,23 +39,6 @@ export enum CurrencyTypeEnum {
...
-    // XCM assets
-    0: 'KSM',
-    1: 'USDT',
-
-    // Stellar assets
-    256: 'XLM',
-}
-
-export const invertedTokenSymbolMap = invert(currencyTokenSymbolMap)
-
```
ebma, line 58: So we don't need this anymore right? I just had a look and I found some leftover TODO comment from myself in the farming.ts file. Should we then remove the TODO comment etc as well?

ebma (commented):
Looks good to me, nice job @gianfra-t  🙏  Just left two minor comments

src/utils/pricing.ts
gianfra-t: Yes, good catch. Otherwise it will probably fail with Pendulum. I will also modify for wherever `WNATIVE` is used so it takes the proper constant.

src/utils/token.ts
gianfra-t: Yes I know the one you are referring to, over [here](https://github.com/pendulum-chain/pendulum-squids/blob/2c892a4cf063ef105dc85ca43b6f91e3ae60d354/src/utils/farming.ts#L211). Indeed we would not use the map so we can remove that too. 

Regarding the comment, I didn't understand it much because `invertedTokenSymbolMap` would map `string-> number` , given the definition of the [map](https://github.com/pendulum-chain/pendulum-squids/blob/2c892a4cf063ef105dc85ca43b6f91e3ae60d354/src/utils/token.ts#L48),  but this would have been accessed with `token0Symbol` which is a `u8` right? Maybe the comment was for an older map?

ebma: I think I remember why we had this inverted map. Some time ago, our `CurrencyId` enum was not using `u8` indices to map to `XCM` currencies but instead we had a separate enum type for foreign currencies. The enum would then have the explicit token names like 'KSM', 'USDT', ... which is why this inverted map made sense.

Since this is not the case anymore, it's safe to remove.

src/utils/pricing.ts
ebma: Thanks for refactoring it into constants 👍 

I thought about the whole price setup again and realized that we need to handle the prices a little differently on Pendulum vs Amplitude. The reason is that the PEN token is listed on a CEX and thus the price is directly provided from the DIA price oracle. The AMPE token is not listed on a CEX so we have to derive the price from the AMPE/KSM Zenlink pool. I committed a small fix for this, please tell me what you think @gianfra-t

gianfra-t: It totally makes sense. Didn't think about getting it from the oracle !

ebma (approved):
Ready for merge IMO

gianfra-t:
@ebma I sent a re-review because for some reason it is preventing the merge, and no changes were done.

ebma:
Ah merging is blocked because the last commit is mine so my approval doesn't count anymore. I'll merge it anyways.

gianfra-t:
Thanks! I will deploy the new versions shortly.

PR #52, opened 2024-02-19
44 migrate to new nabla contracts

### LINKED ISSUE #44: Migrate to new Nabla contracts
The Nabla contracts were changed in [this PR](https://github.com/0xamberhq/contracts/pull/65) and their internal state and abi changed.

The squids should be updated to index the new Nabla contracts.

TorstenStueber:
Closes #44 

This PR implements the indexer changes for Nabla.

The changes are inspired by the according changes of the Ethereum based indexer for Nabla (see this PR: https://github.com/0xamberhq/subgraph-mumbai/pull/11).

However, there are quite a lot of necessary differences that mainly are caused by the fact that the Nabla contract addresses are hard coded in the subgraph indexer whereas we want to avoid hard coded address.

For that reason we need to reliably recognize what contracts are Nabla contracts. In the past we merely probed some methods to heuristically identify whether a contract event is caused by a specific contract type (router, swap pool, backstop pool, etc). However, this proved to be very unreliable and the indexer generated many errors.

Instead we decided to monitor `Contracts.Instantiated` events (see [here](https://github.com/pendulum-chain/pendulum-squids/pull/52/files#diff-7199f058b35aa392107dbd32106c45331e7f2fa9fc6de120e05e78d9db49f0faR324)) and then check the hash of the event causing the event and compare it with the contract metadata to correctly identify the type of contract.

Other differences are to avoid some vulnerabilities that are in the subgraph indexer code:
- correctly identify swap pools that have been added to routers (see [this message](https://satoshipay.slack.com/archives/C03DXSTBWNQ/p1706530838671829))
- correctly assign swap pools for specific tokens (see https://github.com/0xamberhq/subgraph-mumbai/issues/12)

## Tests
I implemented a test suite in wasm deploy (see this PR: https://github.com/pendulum-chain/wasm-deploy/pull/40).

This test suite required to run the wasm deploy test runner and foucoco-standalone locally. This also required to run the indexer (and the graphql server) locally. For that reason I added another configuration option `local` (otherwise subsquid will lookup the configuration for one of our chains in its archive and expects a specific genesis block – which will be different in foucoco-standalone). **I am not sure whether this is the right approach**


src/mappings/farming/handle.ts
```diff
@@ -13,14 +13,11 @@ import {
     formatFarmingPoolEditedEvent,
     formatFarmingPoolKilledEvent,
     formatFarmingPoolResetEvent,
-    formatFarmingRetireLimitSetEvent,
```
TorstenStueber, line 16: These were very unnecessary imports.

src/mappings/token.ts
```diff
@@ -146,7 +146,7 @@ export async function handleTokenDeposited(ctx: EventHandlerContext) {
         currencyId: currencyId,
     })
 
-    ctx.store.save(tokenDeposit)
+    await ctx.store.save(tokenDeposit)
```
TorstenStueber: These were some errors in existing code. We did not await these calls.

src/config.ts
```diff
@@ -1,9 +1,9 @@
 import { ProcessorConfig } from './types'
 import { lookupArchive } from '@subsquid/archive-registry'
 import axios from 'axios'
-export type Network = 'foucoco' | 'amplitude' | 'pendulum'
-export const network: Network =
-    <'foucoco' | 'amplitude' | 'pendulum'>process.env.NETWORK || 'amplitude'
+export type Network = 'foucoco' | 'amplitude' | 'pendulum' | 'local'
+export const network = (process.env.NETWORK as Network) || 'amplitude'
```
TorstenStueber: Alternatively, we could define `network` to be `foucoco` if `process.env.NETWORK` is `local` and only check for equality if `process.env.NETWORK` with `local` when selecting a `ProcessorConfig` [below](https://github.com/pendulum-chain/pendulum-squids/pull/52/files#diff-c3095d5010e65c52737a98a5d618ea24049ebe90c8470752426081d70ed6e012R51).

That way we don't need to change any of the other code that checks the value of the variable `network`.

src/mappings/nabla/handleEvent.ts
```diff
@@ -0,0 +1,67 @@
...
+export async function handleContractEvent(ctx: EventHandlerContext) {
+    const contractSs58Address = hexToSs58(ctx.event.args.contract)
+
+    const [maybeBackstopPool, maybeRouter, maybeSwapPool] = await Promise.all([
+        getBackstopPool(ctx, contractSs58Address),
+        getRouter(ctx, contractSs58Address),
+        getSwapPool(ctx, contractSs58Address),
+    ])
+
+    switch (true) {
```
TorstenStueber, line 54: Unorthodox but clean.

squid-amplitude.yaml
```diff
@@ -1,6 +1,6 @@
 manifestVersion: subsquid.io/v0.1
 name: amplitude-squid
-version: 15
```
TorstenStueber, line 3: There is a conflict with the version bumps from [this PR](https://github.com/pendulum-chain/pendulum-squids/pull/51/files). However, only the Amplitude Squid version 16 has been deployed so far. We probably need to update the version definitions here once all squids from PR #51 have been deployed.

src/mappings/nabla/handleEvent.ts
```diff
@@ -0,0 +1,67 @@
...
+} from './creation'
+import { createBackstopPool, createRouter, createSwapPool } from './creation'
+import { handleBackstopPoolEvent } from './backstopPoolEventHandler'
+import { handleRouterEvent } from './routerEventHandler'
+import { handleSwapPoolEvent } from './swapPoolEventHandler'
+import { hexToSs58 } from './addresses'
+
+export async function handleContractInstantiated(ctx: EventHandlerContext) {
+    const contractHexAddress = ctx.event.args.contract
+    if (contractHexAddress === ZERO_ADDRESS) {
```
gianfra-t, line 19: In which scenario can we end up with the `ZERO_ADDRESS` address for the contract that we want to ignore?

src/mappings/nabla/routerEventHandler.ts
```diff
@@ -0,0 +1,150 @@
...
+        swapPool.token.id
+    )
+
+    if (registeredSwapPool) {
+        if (registeredSwapPool.id === swapPool.id) {
+            // this swap pool is already registered, no action required
+            return
+        }
+
+        registeredSwapPool.router = null
```
gianfra-t, line 124: Here we disconnect the relation between the old swap pool and the router. But what is the reason we want to maintain the old pool in storage? Is it for record purposes only?

src/mappings/nabla/swapPoolEventHandler.ts
```diff
@@ -0,0 +1,241 @@
...
+}
+
+export async function handleBackstopDrain(
+    ctx: EventHandlerContext,
+    swapPool: SwapPool
+) {
+    await updateSwapPoolCoverageAndSupply(ctx, swapPool)
+    await ctx.store.save(swapPool)
+
+    if (swapPool.backstop) {
```
gianfra-t, line 78: Shouldn't always a swap pool have an associated backstop pool on the system? If that is the case, then we could perhaps log an error, otherwise its fine to ignore.

src/mappings/token.ts
gianfra-t: This may not have been a problem since it seems that the deposit entity was not used in the remaining part of the function, but awaiting is also fine.

src/config.ts
gianfra-t: Since we are also using all the `foucoco` events/storage decoders then yes it makes sense!

gianfra-t (commented):
Looks very good! I really like the idea of comparing the hashes of the contract code to identify them.
I just left some comments (questions really).

src/mappings/nabla/routerEventHandler.ts
```diff
@@ -0,0 +1,150 @@
...
+) {
+    const swapPoolSs58Address = hexToSs58(event.pool)
+    const swapPool = await getSwapPool(ctx, swapPoolSs58Address)
+
+    if (swapPool === undefined) {
+        // this is a non-standard or malicious swap pool, ignore
+        return
+    }
+
+    const token = await getOrCreateNablaToken(ctx, ss58ToHex(swapPool.token.id))
```
bogdanS98: Shouldn't this variable be passed as a parameter when calling `getSwapPoolsOfRouterForToken` below? Otherwise, it looks like it's not used at all.

src/mappings/nabla/creation.ts
```diff
@@ -0,0 +1,220 @@
...
+    return nablaToken
+}
+
+export async function createSwapPool(
+    ctx: EventHandlerContext,
+    hexAddress: string
+) {
+    const ss58Address = hexToSs58(hexAddress)
+    if (await getSwapPool(ctx, ss58Address)) {
+        // Unexpected error: swap pool already exists at that address
```
ebma, line 145: It could be worth adding `console.log()` statements to all the early returns if a contract already exists at an address. This is an unlikely event but having the log statements would facilitate debugging issues related to this.

ebma:
@TorstenStueber I pushed a new [commit](https://github.com/pendulum-chain/pendulum-squids/pull/52/commits/4ad7c45197b65b58ec9c9bcd59fdad0b65215464) that updates the ABI files of the Nabla contract to the ones compiled with a later version of Solang ie. v0.3.3 instead of v0.3.2. Please be aware that this now requires you to compile your Nabla contracts with the same version of Solang before using wasm-deploy to deploy them again. 

I only switched to [v0.3.3.](https://github.com/hyperledger/solang/releases/tag/v0.3.3) because it was the latest version available of Solang when I started testing everything one week ago. We can also roll it back if you want, but I guess it's good to stay up-to-date with the latest versions.

TorstenStueber:
@ebma we would either need to roll it back or deploy Nabla again on Foucoco. At least the current Nabla instance we use for all testing would not work with this version of the indexer anymore.

ebma:
Then let's deploy a new version of Nabla on Foucoco again or do you think this causes too much extra effort? AFAIK we'd would just need to change some hard-coded addresses in the portal and the rest can stay the same.

src/config.ts
TorstenStueber: I changed the logic: https://github.com/pendulum-chain/pendulum-squids/pull/52/commits/84d220c8ce2d0fa32da16822f50446ba8db77843

src/mappings/nabla/handleEvent.ts
TorstenStueber: It probably never happens. I added this to be defensive. If for strange reason this would be the zero address, then the following code would fail (add an entry to the database with id 0).

src/mappings/nabla/routerEventHandler.ts
TorstenStueber: I think I followed the original logic of the Ethereum based indexer of the Nabla team. It might be better to keep it as this is a scenario that would happen whenever we need to adapt the slippage curve (e.g., we see that slippage is too low or too high in practive). The slippage curve of a swap pool is immutable by design and one would deploy and attach a new swap pool for the same token but a different slippage curve. After that liquidity providers would need to withdraw liquidity from the old pool and deposit into the new pool.

We don't have a UI for this yet (makes sense to only implement once we need it) but for that case we would need to keep the old swap pool in the indexer.

src/mappings/nabla/swapPoolEventHandler.ts
TorstenStueber: The case that there is no backstop pool should probably never occur but I did not just want to force a type assertion here.

src/mappings/nabla/routerEventHandler.ts
TorstenStueber: You are right. Actually the result is not required and it just matters that the function is called. I changed it and [opened a new ticket](https://github.com/pendulum-chain/pendulum-squids/issues/63) to avoid these errors in the future.

TorstenStueber:
@pendulum-chain/devs I addresses the comments from earlier reviews.

This PR can now be properly reviewed.

ebma (approved):
Looks good to me 👍

src/mappings/nabla/creation.ts
TorstenStueber: It would be certainly helpful to make this more visible.

Is this the right pattern to use `console.log()` in the indexer code? Would we be able to see that in the subsquid log views? It seems that there are no other `console.log()` in event processing code.

ebma: Yes, you can use `console.log()` and we will see it, no problem. We are already using it for errors during processing e.g. [here](https://github.com/pendulum-chain/pendulum-squids/blob/993d2f9f2eed0ff21c319c03358b8af634eec5f2/src/processor.ts#L341).

TorstenStueber: I added some log statements: https://github.com/pendulum-chain/pendulum-squids/pull/52/commits/8a543ff3b98523cb9c628724020f5f7568d9f57d

ebma: Thanks 👍

bogdanS98 (approved):
Looks good to me too 👍🏼

ebma:
Let's merge @TorstenStueber?

gianfra-t (approved):
I checked the last commits since the previous review and looks all good also!

PR #56, opened 2024-03-14
Fix for sorting pairs before querying the storage

### LINKED ISSUE #57: Prices and amount not updated on Zenlink UI for PEN/DOT pool
## Context
After performing some swaps the prices on zenlink UI for both PEN and DOT dont change neither do the amount of tokens in the pool. Expected to change the amount of tokens and the price of the tokens in the pool

## Expected outcome
* Prices of the tokens should update after swapping
* Amount of the tokens should update after swapping

## Reference
image

gianfra-t:
A bug was identified where the [lookup](https://github.com/pendulum-chain/pendulum-squids/blob/1a942272345091fdebf32be6ef5d692212de7aee/src/entities/pair.ts#L33) of the pair given an event swap was not returning the pair, and therefore not indexing the swap.

 Since `getPairAssetIdFromAssets` calls the [storage](https://github.com/pendulum-chain/pendulum-squids/blob/1a942272345091fdebf32be6ef5d692212de7aee/src/utils/token.ts#L246), at this point we need to sort the assets the same way the zenlink pallet does.

The rest of the changes are to adapt the code so that even when we have to reorder the pairs, we are consistent with the `asset0`, `asset1` naming and convention. For example, `amount0In` should always correspond to the input amount into the swap of the first asset **after they are sorted**.

Closes #57.

### DIFF
--- a/squid-amplitude.yaml
+++ b/squid-amplitude.yaml
@@ -1,6 +1,6 @@
 manifestVersion: subsquid.io/v0.1
 name: amplitude-squid
-version: 16
+version: 17
 description: 'Amplitude Kusama Squid'
 build:
 deploy:
--- a/squid-amplitude.yaml
+++ b/squid-amplitude.yaml
@@ -1,6 +1,6 @@
 manifestVersion: subsquid.io/v0.1
 name: amplitude-squid
-version: 17
+version: 18
 description: 'Amplitude Kusama Squid'
 build:
 deploy:
--- a/squid-foucoco.yaml
+++ b/squid-foucoco.yaml
@@ -1,6 +1,6 @@
 manifestVersion: subsquid.io/v0.1
 name: foucoco-squid
-version: 17
+version: 18
 description: 'Foucoco Squid'
 build:
 deploy:
--- a/squid-foucoco.yaml
+++ b/squid-foucoco.yaml
@@ -1,6 +1,6 @@
 manifestVersion: subsquid.io/v0.1
 name: foucoco-squid
-version: 16
+version: 17
 description: 'Foucoco Squid'
 build:
 deploy:
--- a/squid-pendulum.yaml
+++ b/squid-pendulum.yaml
@@ -1,6 +1,6 @@
 manifestVersion: subsquid.io/v0.1
 name: pendulum-squid
-version: 17
+version: 18
 description: 'Pendulum Squid'
 build:
 deploy:
--- a/src/mappings/protocol.ts
+++ b/src/mappings/protocol.ts
@@ -11,7 +11,6 @@ import {
     getTokenBalance,
     getTokenBurned,
     zenlinkAssetIdToCurrencyId,
-    sortAndCheckIfSwitched,
 } from '../utils/token'
 import { ZERO_BD } from '../constants'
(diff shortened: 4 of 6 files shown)


ebma (approved):
Nice catch with the sorting 👍 

I saw that we already have an existing function that is used in the `utils/farming.ts` file for example so I removed the other one. I also double-checked with the bifrost squid implementation [here](https://github.com/zenlinkpro/bifrost-kusama-squid/blob/6f6358b7c8491038d17a8e35903e3df05114c468/src/mappings/protocol.ts#L278) to make sure we don't miss anything.

gianfra-t:
I missed the existing sorting function, thanks! 
I am not entirely sure about the last changes. I know that the previous condition may not be pretty, but [here](https://github.com/pendulum-chain/pendulum-squids/pull/56/files#diff-6bfc6880709344c4d89f4b711317a45fa1ba101d3baf282206956860c375a04cR349) we are always assuming that `amount0Out` will be 0, similarly with `amount1In`. This will work only for the cases were the swap is already ordered.

I am actually surprised to see that this is the case with the bifrost squid implementation. Perhaps for them, they can query a different pair for a switched order, which would make sense to hold this assumption. What do you think @ebma ?

ebma:
I think it works because the `amount0In` and `amount0Out` (same for amount1) are mutually exclusive, meaning only one of them can have a value and the other will be `0`. If you check [here](https://github.com/pendulum-chain/Zenlink-DEX-Module/blob/7662acbff71cea3b35db918ceb9370e8929be991/zenlink-protocol/src/swap/mod.rs#L269-L270), when calling the extrinsic you specify `amount_in` and `amount_out` and depending on the order of assets in `path` it will either be `amount_in` for asset 0 or for asset 1, with the remaining asset being used for `amount_out`. 

With [this](https://github.com/pendulum-chain/pendulum-squids/blob/4997d44b1bb84e5a423025b948885b15386a3ef5/src/mappings/protocol.ts#L345) check we then determine which amount should be used for which token, so it should work. You could argue that the declaration of `amount0Out` and `amount1In` in our code is now useless as they will always be `0`. Though I guess it's okay to keep it for illustrative purposes. We can remove the useless variables but I think the logic overall should work like this anyways.

gianfra-t:
Okay so it seems then that like you say, it still will works. But it is a little bit confusing to me. If we look at the swap events for these two alternatives:
image
image
we can see that the first one is telling us that `DOT` was the asset in, and `PEN` the asset out, which we know was the case from the actual transaction. This is with the version 18 uploaded yesterday.

On the other hand, in the latest iteration this is reversed. I checked in the remaining parts of the code and the only variable used is the total amount, so it doesn't break anything but the record itself.

ebma:
Hmm you are totally right, thanks for elaborating on this in more detail. 

I changed it back to something that should work, please have another look @gianfra-t.

gianfra-t:
Alright, seems good! Thanks.  I will bump the other versions because it should affect them also.

ebma:
Let's bump all squids to the same version (in this case 18). Then it's easier to see in the cloud app which squid has which features.

ebma (approved):

PR #59, opened 2024-03-18
Await database operations, add eslint rule

### LINKED ISSUE #55: Await database operations
The squid on Amplitude is frequently facing errors with

```
too late to perform db updates, make sure you haven't forgot to await on db query
```

which probably happens because we don't `await` the database operations in all functions that interact with the squid `store` object.

## TODO 
Use `await` in functions that interact with the `ctx.store`, e.g. [here](https://github.com/pendulum-chain/pendulum-squids/blob/1a942272345091fdebf32be6ef5d692212de7aee/src/mappings/nabla.ts#L197).
To make sure we don't miss this we can think about adding [this](https://eslint.org/docs/latest/rules/require-await) ESlint rule.

bogdanS98:
Closes #55.

This PR adds and configures eslint. The configuration extends the default recommended rules but two rules were deactivated for convenience while other rules to warn about missing `await` statements were added. 
This PR also fixes all the warnings/errors thrown by ESlint for this configuration.

Note that the linting scripts are not run automatically but have to be run manually.


ebma (commented):
Nice 👍 

Let's make it a habit of always bumping squids to the same version. Should we write down this rule in the README or what do you think would be the best way to make this clear @bogdanS98 @gianfra-t? 

What's the remaining problem with eslint you are facing, need any help with that?

gianfra-t:
I think it makes sense to put it in the README, somewhere near the commands to deploy probably.

ebma:
I finished setting up ESLint and fixed the warnings/errors. 

@gianfra-t and @bogdanS98 do you think we should run the `lint` script automatically on every commit like we do with `prettier` or is this too much? I decided not to make it execute automatically.

gianfra-t:
I think is best to leave it for manual check, or at most a push hook like we did with cargo clippy. But I may be biased since I don't use `lint` much...

gianfra-t (approved):

PR #61, opened 2024-03-26
Update types foucoco amplitude v0942

### LINKED ISSUE #60: Regenerate events and storage types after Amplitude, Foucoco and Pendulum runtime upgrades
After the upgrade to `0.9.42`, we need to regenerate the types and upload a new squid version for every chain.

gianfra-t:
Closes #60.

### Steps
- [x] Regenerate types and deploy Foucoco squid.
- [x] Regenerate types and deploy Amplitude squid.
- [x] Regenerate types and deploy Pendulum squid.

Note: all new squid versions should be bumped to `v20`.


ebma:
Nice 👍  can we merge https://github.com/pendulum-chain/pendulum-squids/pull/59 first? If we deploy these squids without it, they would again miss the `await` statements and I'll get spammed with emails that our squid missed to wait for database operations.

gianfra-t:
Sure you are right. I will merge it but only bump the version here if that's okay.

ebma:
Sure, let's merge it into this one and go for v20 👍

gianfra-t:
This is an odd case because the Pendulum squid should also be bumped to `v20` since it has some changes, yet we cannot upgrade the types still. Upgrading it to `v20` now would be confusing since it would be 1 version ahead later.

Perhaps we could leave this PR open until we can upgrade the types for Pendulum and deploy all of them. In the meantime, I deployed Amplitude/Foucoco.

ebma:
> Perhaps we could leave this PR open until we can upgrade the types for Pendulum and deploy all of them.

Good idea 👍 To make it more transparent please add a new TODO checklist item to the description of the PR that mentions we still need to update the types on Pendulum.

ebma (approved):
Let's merge now that all types are up-to-date.

PR #67, opened 2024-06-13
Reset websocket connection if syncing fails

### LINKED ISSUE #65: Reset websocket connection if syncing fails
We learned that the indexer might just stop syncing if the websocket connection to our RPC node hangs.

This can be avoided by setting a timeout for the connection. If the indexer recognizes that it cannot sync anymore, it will then restart the connection.

[Here is documentation](https://docs.subsquid.io/sdk/reference/processors/substrate-batch/general/#set-rpc-data-ingestion-settings) about the API that will allow to set this timeout.

Also see this screenshot:

Screenshot 2024-06-05 at 08 25 11

ebma:
I already deployed the new squids v23 in the cloud so we can see if they are able to catch up. 

Closes #65.

### DIFF
--- a/squid-amplitude.yaml
+++ b/squid-amplitude.yaml
@@ -1,6 +1,6 @@
 manifestVersion: subsquid.io/v0.1
 name: amplitude-squid
-version: 22
+version: 23
 description: 'Amplitude Kusama Squid'
 build:
 deploy:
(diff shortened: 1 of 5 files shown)


gianfra-t (approved):
There should be a default timeout implemented for this!

ebma:
Yes 😅 
I checked the v23 squids and they are all functional. I bumped them to be the production endpoints and will delete the others around Monday probably.

PR #68, opened 2024-08-02
Create local archive endpoint to allow updating indexers before enactment of runtime upgrades

### LINKED ISSUE #64: Create local archive endpoint to allow updating indexers before enactment of runtime upgrades
## Context

In https://github.com/pendulum-chain/tasks/issues/186 we did some research how we can improve the indexer downtime whenever a runtime upgrade happens. In the past we noticed that the metadata changes introduced by runtime upgrades make our indexers unable to decode some of the events properly. 

We now want to implement the agreed outcome of the research task. The idea is the [following](https://github.com/pendulum-chain/tasks/issues/186#issuecomment-2061382585):

#### Local archive endpoint
We create a new command or script that:
- Assumes you have the new `runtime.wasm` file somewhere in the local directory
- The script calls the [subwasm](https://github.com/chevdor/subwasm/) tool on `runtime.wasm` to extract the metadata from it. The metadata is then stored in a file `runtime.metadata`
- The script starts a new local HTTP server that:
   - Queries the respective production metadata archive, e.g. https://v2.archive.subsquid.io/metadata/foucoco
   - Reads the content of the local `runtime.metadata` file we created earlier and creates a new JSON object of the form `{"blockNumber":0,"blockHash":"xxx","specName":"foucoco","specVersion":<the_new_spec_version>,"metadata":"<content_of_runtime.metadata>"}` with it
   - Serves GET requests with the concatenation of the original metadata archive response + the custom JSON

## TODO

- Implement the server for the local archive endpoint. The server can be implemented using [express](https://expressjs.com/) or just the `http-server` package. It should live in an extra folder. It does not need to live in the `src` folder with the rest of the squid files so we can reduce the overhead by using Javascript over Typescript.
- Add a command to [commands.json](https://github.com/pendulum-chain/pendulum-squids/blob/main/commands.json) that first runs subwasm to extract the metadata and then starts the local server.
- (Optional) Find a way to automatically run the respective `typegen-{network}` command to regenerate the types. This might be tricky as the typegen commands need to know the network they are used for and we need to adjust the contained specVersions endpoint eg. [here](https://github.com/pendulum-chain/pendulum-squids/blob/993d2f9f2eed0ff21c319c03358b8af634eec5f2/typegen-foucoco.json#L3).

bogdanS98:
Closes #64.

### Notes: 

- Foucoco runtime and metadata files will be removed after review as they are only used for testing. The provided Foucoco runtime has spec version 19 and contains an extra dummy pallet compared to the latest live runtime.
- `subwasm` tool is a prerequisite and must be installed on the local machine in order for the script to work.

### How it works

This PR introduces a local archive endpoint and the logic for automated type regeneration based on a new custom metadata file.

When running `sqd generate-types-from-runtime-binary <runtime_name> <path_to_runtime_binary>`, the following will happen (in order):

- `subwasm` is ran against a local runtime binary at `<path_to_runtime_binary>` which produces a `<runtime_name>.metadata` file containing the hex+scale encoded metadata.
- The corresponding `typegen-<runtime_name>.json` is updated to contain the local archive endpoint URL on the `specVersions` field i.e. `http://localhost:3000`.
- Local archive endpoint is started at `http://localhost:3000`.
- `sqd typegen:<runtime_name>` command is ran which in turn will spawn a process that will make a request to the URL found in the `specVersions` field of `typegen-<runtime_name>.json`.
- Server receives request, fetches the original metadata file from `https://v2.archive.subsquid.io/metadata/<runtime_name>`, increments the last `specVersion` found in the original file, adds it to the new custom metadata file along with the `subwasm` generated metadata and then serves the concatenated original metadata + custom metadata files.
- After types are generated successfully based on the new metadata file, `typegen-<runtime_name>.json` is updated again with the old URL value for `specVersions` field i.e. `https://v2.archive.subsquid.io/metadata/<runtime_name>` and server gracefully shuts down.


local-archive-endpoint/foucoco_runtime.compact.compressed.wasm
ebma, line 1: Do we need to check this file into the repo?

local-archive-endpoint/runtime.metadata
ebma, line 1: Can also be removed?

commands.json
```diff
@@ -130,6 +130,10 @@
         "open": {
             "description": "Open a local browser window",
             "cmd": ["npx", "--yes", "opener"]
+        },
+        "pre-runtime-upgrade": {
+            "description": "Pre-runtime upgrade type generation for a specific runtime",
+            "cmd": ["npm", "run", "pre-runtime-upgrade"]
```
ebma: I think I would phrase the command a little differently.

```suggestion
        "generate-types-from-runtime-binary": {
            "description": "Generates the respective type files for a provided parachain runtime binary",
            "cmd": ["npm", "run", "generate-types-from-runtime-binary"]
```

local-archive-endpoint/server.js
```diff
@@ -0,0 +1,182 @@
...
+        }
+    } catch (error) {
+        console.error(`Error running subwasm or typegen: ${error.message}`)
+        process.exit(1)
+    }
+}
+
+const runtimeParam = process.argv[2]
+if (!runtimeParam) {
+    console.error('Usage: sqd pre-runtime-upgrade <runtime_name>')
```
ebma: Maybe it's better to pass the path to the file instead of making assumptions about where it's stored and what it's called. 
```suggestion
    console.error('Usage: sqd pre-runtime-upgrade <runtime_name> <path_to_runtime>')
```

package.json
```diff
@@ -12,7 +12,8 @@
         "query-node:start": "squid-graphql-server --subscriptions --max-response-size 10000 --dumb-cache in-memory --dumb-cache-ttl 1000 --dumb-cache-size 100 --dumb-cache-max-age 1000",
         "prepare": "husky install",
         "lint": "eslint . --ignore-path .eslintignore --ext .ts",
-        "lint:fix": "eslint . --ignore-path .eslintignore --ext .ts --fix"
+        "lint:fix": "eslint . --ignore-path .eslintignore --ext .ts --fix",
+        "pre-runtime-upgrade": "node local-archive-endpoint/server.js"
```
ebma:
```suggestion
        "generate-types-from-runtime-binary": "node local-archive-endpoint/server.js"
```

ebma (commented):
I didn't run it locally yet.

local-archive-endpoint/foucoco_runtime.compact.compressed.wasm
bogdanS98: I only left it so that it's easier for reviewers to run the script locally and not have to provide a runtime wasm. I will remove it after review in case anyone wants to test the code.

ebma:
Let's also add a description how to use the new commands to the README.

ebma (approved):
I tested it locally now and it works great 👌 I just added some smaller changes but I think it's ready for merge.

gianfra-t (approved):
Looks good!! Thanks @bogdanS98.

PR #69, opened 2024-09-10
Increase decimal precision and head timeout

ebma:
This PR increases the precision of the numbers stored by the squid from 6 to 10 decimals. We do this to increase the precision of the prices shown in the Zenlink UI. The problem is that the AMPE price is currently very low with just a fraction of a cent. Since for the Zenlink entities, prices are derived relative to the native currency, it's important to increase the number of decimals so that the AMPE price is represented with higher precision and less rounding occurs during the calculations of the other prices. 

The `newHeadTimeoutMs` parameter is increased from 1 minute to 5 minutes. We were repeatedly seeing errors in the squid logs that caused the squid to restart. I identified that these errors happen because of the `newHeadTimeoutMs` parameter.  We use the `newHeadTimeoutMs` parameter to make the squid automatically restart the RPC connection in case it gets stuck. Setting it to 1 minute is too low, considering the sometimes high block latency of our runtimes which caused unnecessary restarts and errors thrown in the squids. 

Besides that, this PR also replaces the deprecated call to `setDataSource` with the proper calls to `setGateway` and `setRpcEndpoint`. The list of supported gateways can be found [here](https://docs.sqd.dev/subsquid-network/reference/substrate-networks/). 

Closes https://github.com/pendulum-chain/tasks/issues/395.


gianfra-t (approved):
LGTM. 

Did you test it that this solves the price issue? Otherwise we can deploy one squid from this branch and test it first.

ebma:
Yes, I tested it. All networks are already running with these new changes and the accuracy improved.

PR #70, opened 2024-09-18
Fix foucoco squid issues

ebma:
Since Foucoco (Paseo) created the genesis block with spec version 18, we needed to regenerate the types and adjust the types array. Otherwise, the types are inconsistent with the spec versions. 

Besides that, we can't use the Foucoco archive gateway for the time being. The subsquid team is currently resetting that endpoint but until it's ready, our Foucoco squid needs to query all the historical data from the RPC node. 

There also was a bug with the `maxHeightPromise`. We use this variable to determine the blocks we want to keep in storage because to reduce the storage requirements of the squid, we only store blocks for the last 24 hours. If the query for the `maxHeightPromise` fails, we previously returned the `MAX_SAFE_INTEGER`. This is bad because it fools our block retention check to believe that the current block height is super far in the future, so the processor will not store _any_ blocks. This is fixed by choosing the 0.


src/mappings/balances.ts
```diff
@@ -1,11 +1,11 @@
...
 import { EventHandlerContext } from '../processor'
 import { Transfer } from '../model'
-import events from '../types/events'
-import { config } from '../config'
+import { config, network } from '../config'
+import { decodeEvent } from '../types/eventsAndStorageSelector'
 
 export async function handleBalanceTransfer(ctx: EventHandlerContext) {
-    const event = events.balances.transfer.v1.decode(ctx.event)
+    const event = decodeEvent(network, ctx, 'balances', 'transfer')
```
gianfra-t, line 8: Good catch, we missed some in the past.

gianfra-t (approved):
LGTM 👍. Sorry I missed this review. 

In this case, you used the old `typegen:foucoco` since the archive already had the new `v18` metadata right?

ebma:
Yes, I used `typegen:foucoco` 👍  Not because the archive already had the metadata (it was down/migrated at the time) but because the v18 runtime was already live and the _RPC/collator node_ already had the metadata. This is a small detail but to clear confusion.

PR #71, opened 2024-10-03
Fix and run eslint.

### LINKED ISSUE #63: Fix eslint
I observed some errors in the code that would have been easy to prevent with improved linting rules (or linting is set up incorrectly).

This is:
* definition of variables that are not used
* (particularly important): forgetting to await the return result of a function (that returns a promise)

## Context
This configuration change will make some errors less likely.

gianfra-t:
#63 

This PR does not include any config changes since the scripts `lint` and `lint:fix` are working for me. I just run `fix`.


ebma (approved):

ebma:
Please don't forget to add reviewers @gianfra-t 😅

ebma:
Actually, I tested this locally and I also get an error
```
$ yarn lint
Invalid option '--ignore-path' - perhaps you meant '--ignore-pattern'?
You're using eslint.config.js, some command line flags are no longer available. Please see https://eslint.org/docs/latest/use/command-line-interface for details.

```

I did some digging and I think we need to migrate to a new format. It's probably easiest if we just start with the configuration from scratch using the steps [here](https://typescript-eslint.io/getting-started/). I already tested it and got 3126 problems (3126 errors, 0 warnings). Maybe we should ignore some of the automatically generated files from eslint as it doesn't really make sense to fix warnings on them as they would need to be fixed again and again. 

Can you please add the migration to the new eslint to this PR @gianfra-t?

gianfra-t:
Okay, I will change it. Strange that I don't get an error 🤔 nor a warning.

.eslintrc.json
```diff
@@ -1,20 +0,0 @@
...
-    },
-    "extends": ["eslint:recommended", "plugin:@typescript-eslint/recommended"],
-    "parser": "@typescript-eslint/parser",
-    "parserOptions": {
-        "ecmaVersion": "latest",
-        "project": "./tsconfig.json"
-    },
-    "plugins": ["@typescript-eslint"],
-    "rules": {
-        "@typescript-eslint/no-unused-vars": "off",
```
gianfra-t, line 15: @pendulum-chain/devs any idea why we had disable this unused variables rule? I enabled it in the new config. But I could be missing something.

eslint.config.mjs
```diff
@@ -0,0 +1,29 @@
...
+    ],
+    languageOptions: {
+        parser: tseslint.parser,
+        parserOptions: {
+            projectService: true,
+            ecmaVersion: 'latest',
+            tsconfigRootDir: './tsconfig.json',
+        },
+    },
+    rules: {
```
gianfra-t, line 23: @pendulum-chain/devs any idea why we had [disabled](https://github.com/pendulum-chain/pendulum-squids/pull/71/files#diff-6884918dc8291219be508e05e28965b958c734def85324f3b53858ea4702090fL15) the `unused variables` rule? I enabled it in the new config here. But I could be missing something.

ebma: I think it's because the automatically generated files yield a lot of the unused variable errors. If you remove 'src/types/**' from the `ignores` field and run lint again, you'll see.

But now that we ignore those files anyway, I think it's good to enable it again 👍

ebma (approved):
Tested locally, works 👍  nice job @gianfra-t

TorstenStueber:
Can confirm that it works for me, too! Great!

PR #72, opened 2024-10-04
Use yarn and resolve dependency conflicts

ebma:
I forgot to change this repository to yarn as well.
Related to https://github.com/pendulum-chain/tasks/issues/402.


gianfra-t (approved):
Thanks @ebma. I actually changed it yesterday but later rolled back since on the `README` it was stated that "`yarn` package manager is not supported". Was that a leftover comment? Or it was still a recommendation?

ebma:
Yeah good point, I removed that as part of this PR. It was part of the original template that the subsquid team created for substrate squids. I think it's because some of their commands explicitly used `npm` but I changed it so that everything should work with yarn. I don't see why it wouldn't work anymore so I would consider it an outdated comment from the past.

PR #73, opened 2024-10-04
Handle Issue and Redeem requests.

### LINKED ISSUE #66: Handle issue/redeem requests in indexer
#### Context
There is a problem with the transaction times displayed in "Spacewalk > Transfers" on the Pendulum portal. The times change every 10-12 seconds due to the automatic table refresh with each new block in the chain. The estimateRequestCreationTime function estimates the transaction time based on a 12-second interval between blocks, but since blocks do not appear exactly every 12 seconds, the estimated transaction time fluctuates.

Issue:
- The transaction time changes from, for example, 10:44:23 to 10:44:42 after an automatic refresh for the same transaction.

#### Requirements
- [ ] Implement the feature to handle issue/redeem requests in the indexer.
- [ ] Ensure the indexer includes the necessary metadata: on-chain data and the creation timestamp of requests.
- [ ] Modify the portal to use this new data from the indexer for consistent timestamps.

gianfra-t:
Issue: #66.

Highlights:
- [Added](https://github.com/pendulum-chain/pendulum-squids/pull/73/files#diff-efc1675187620595b0844197a25c35eac9b6df752f9182e5cffddee6f27a8489R856) the schema for `Vault`, `IssueRequest` and `RedeemRequest`.
- [Mappings](https://github.com/pendulum-chain/pendulum-squids/pull/73/files#diff-360952853cf1a28e9d91b37e26a515116b5eaee233caf54c8983ed19aad67743R12) for the respective events.

Approach: Upon each event received, we query the storage value for the corresponding redeem/issue request and update the indexer's entry based on it. `Vault` entity is created (if needed) when receiving an issue event.


gianfra-t:
For the portal:

The relevant query to the indexer that now would replace the timestamp [estimation](https://github.com/pendulum-chain/portal/blob/fe42fefa316f5de76e2932cd535ebf6a221c09a1/src/pages/spacewalk/transactions/index.tsx#L62) would be something like:
```
query GetIssueRequestTimestamp {
  issueRequests(where: {requester_eq: "account"}) {
    issueId
    timestamp
  }
}
```

For future improvements, we have almost all the data to avoid [fetching all entries](https://github.com/pendulum-chain/portal/blob/fe42fefa316f5de76e2932cd535ebf6a221c09a1/src/pages/spacewalk/transactions/index.tsx#L44) and filtering, except for the status which of course cannot be inferred from the initial event.

For that, we would have to add listeners for `Cancel, Execute` etc. and update the storage accordingly.

src/model/generated/backstopPool.model.ts
```diff
@@ -21,13 +24,13 @@ export class BackstopPool {
     @PrimaryColumn_()
     id!: string
 
-    @Column_('text', { nullable: false })
```
ebma, line 24: I assume all of these changes in the model are from a prior dependency change (not included in this particular PR) but we never ran the model generation again (?) Looks like an improvement to me though.

src/mappings/spacewalk.ts
```diff
@@ -0,0 +1,57 @@
...
+    await ctx.store.save(redeemRequest)
+}
+
+function parseVaultId(vaultId: any): VaultIdFlat {
+    return {
+        accountId: vaultId.accountId,
+        collateral: beautifyCurrencyIdString(vaultId.currencies.collateral),
+        wrapped: beautifyCurrencyIdString(vaultId.currencies.wrapped),
+    }
+}
```
ebma: Let's move `parseVaultId()` to a better location. I would group it with a new function that derives the accountId for `src/entities/vault.ts` 
```
function deriveVaultIdString(vaultId: VaultIdFlat) {
   return `${vaultId.accountId}-${vaultId.collateral}-${vaultId.wrapped}`
}
```

Maybe put it to a file `src/utils/vault.ts`.

src/types/pendulum/v12.ts
ebma, line 1: Did you only regenerate the types for Pendulum? Let's also do it for Foucoco and Amplitude.

ebma (commented):
The implementation is nice, great job @gianfra-t!

I thought about the implications for the portal and I would appreciate it if we can actually have all available attributes on the indexer. This means, we should add `opentime`, `period`, `griefingCollateral`, `asset`, `fee` and `status` to issue requests and `opentime`, `period`, `fee`, `asset`, `premium` and `status` to redeem requests. I know that the asset is inferrable by the vault linked to each request so we can keep that deduplicated if you prefer it this way. However, I think for the portal to really profit from the additions to the indexer, it would be very nice if it can solely fetch all relevant data from it. Without these fields, the portal would need to query both the indexer and onchain data to show some of the info in the dialogs which can be avoided if we do it this way.

I know that not all these fields are included in the events but how about we change it so that the indexer queries the on-chain data of the `issue_id` or `redeem_id` from the storage item on either `RequestXX`, `ExecuteXX` or `CancelXX` (+ `LiquidationRedeem`, `MintTokensForReimbursedRedeem` and `SelfRedeem`) instead of filling it with the event data? We know that this is possible since we do it similarly for the smart contract data.

src/model/generated/backstopPool.model.ts
gianfra-t: Yes most likely because of an update and running the generation from the schema again. I only added those 3 instances I referenced in the description.

src/types/pendulum/v12.ts
gianfra-t: Yup, to avoid even more file changes. But is good that you reminded me to do this at the end!

gianfra-t:
@ebma sure, I agree that now that we are changing this we might as well do it once. But I have a few questions to avoid going overboard with what event data we parse, especially for those entities where the states are dependent of many events and can be difficult to debug.

I thought the only missing element that the Portal needed to query was the `status` field, that I commented earlier. That can be easily added by handling `Cancel` or `Execute` events. (and probably `LiquidationRedeem` also).
But what about the others? I quickly searched on the portal code and I see that `griefingCollateral` is also used, ~~so this one should be updated to and queried from the new `vault` entity, right?~~ Nevermind this last part, the griefing collateral is associated with a request event only.

ebma:
The thing is that we also use other attribute of the requests, e.g. the `opentime` and `period` to show a countdown timer of the remaining time that the user has for the transfer on Stellar, see [here](https://github.com/pendulum-chain/portal/blob/fe42fefa316f5de76e2932cd535ebf6a221c09a1/src/components/TransferCountdown/index.tsx#L27-L28). These parameters are never included in any event of the events so I proposed to just check the on-chain storage items. We can explicitly query from a storage item at any time with code similar to [this](https://github.com/pendulum-chain/pendulum-squids/blob/db87fdd42d5935a8c6020a562dfc2310edf4320a/src/utils/farming.ts#L104). I think performance-wise it should be fine if we do this considering that there are not that many Spacewalk-related events. At least in comparison to other events that are emitted on every block.

gianfra-t:
Great. I checked the `period` parameter and we can totally fetch it from the storage when the event is received, and add it to the instance. For `opentime` I was thinking of using the current block to save a query, but that would be incorrect.

ebma:
All of these fields are available on the IssueRequest struct that is stored on chain, see [here](https://github.com/pendulum-chain/spacewalk/blob/7c7989875e95e1cfe3b0aeba9bb6af01d9e33e58/primitives/src/lib.rs#L172), so I think all we need to do is query the on-chain storage item and that's it, no need to look at anything else 😬 They are populated when the issue is first created etc.

src/mappings/token.ts
```diff
@@ -62,67 +62,59 @@ function deriveStellarPublicKeyFromHex(issuer: string) {
     return StrKey.encodeEd25519PublicKey(buffer)
 }
 
-function beautifyCurrencyIdString(event: any) {
-    let currencyId = ''
-
-    switch (event.currencyId.__kind) {
+export function beautifyCurrencyIdString(currencyId: any) {
+    switch (currencyId.__kind) {
```
TorstenStueber, line 66: We save a line of code here but I think it is not super clean (rather unexpected and surprising) to overwrite the argument variable and to even overwrite it in a way that its type changes.

This makes the code in this function brittle as it is not obvious where the variable still has its old value and where it already has its new value.

Instead: replace the `break` statements by returning the result value directly in each case.

src/mappings/spacewalk.ts
```diff
@@ -0,0 +1,184 @@
...
+import { getOrCreateVault } from '../entities/vault'
+import { parseVaultId } from '../utils/vault'
+import { getVersionedStorage } from '../types/eventsAndStorageSelector'
+import { network } from '../config'
+import { IssueRequestStatus } from '../model/generated/_issueRequestStatus'
+import { RedeemRequestStatus } from '../model/generated/_redeemRequestStatus'
+
+export async function handleIssueRequest(ctx: EventHandlerContext) {
+    const { args } = ctx.event
+    const {
```
TorstenStueber: Is the `asset` argument not interesting to us (or is it that we can derive it from the vault id)?

src/mappings/spacewalk.ts
```diff
@@ -0,0 +1,184 @@
...
+    }
+
+    // we update on this event the slashed collateral if the amount sent was not enough. The amount is updated on the executed event.
+    issueRequest.slashedCollateral = confiscatedGriefingCollateral
+    await ctx.store.save(issueRequest)
+}
+
+export async function handleRedeemRequest(ctx: EventHandlerContext) {
+    const { args } = ctx.event
+    const { redeemId, redeemer, amount, vaultId, fee, premium, transferFee } =
```
TorstenStueber: Likewise, `stellarAddress and `asset` not interesting to us?

src/mappings/spacewalk.ts
```diff
@@ -0,0 +1,184 @@
...
+    if (redeemRequest == null) {
+        throw new Error(
+            'Redeem request MUST exists on the database when handling cancel event'
+        )
+    }
+
+    if (status == 'Reimbursed') {
+        redeemRequest.status = RedeemRequestStatus.REIMBURSED
+        // Todo, not sure how this status is represented.
+    } else if (status == 'Reimbursed(true)?') {
```
TorstenStueber: Is the stringified status really `Reimbursed` or `Reimbursed(true)?` (with the question mark) when the Rust enum defines it as `Reimbursed(bool)`?

src/mappings/spacewalk.ts
gianfra-t: I proposed we get it from the vault entity to avoid duplication. From `vaultIdFlat.wrapped`, if we ever need it.

gianfra-t: I can add the stellar address though, for the future.

src/mappings/spacewalk.ts
gianfra-t: No it isn't !! I need to look for an event to see the actual property. Sadly `Pendulum` doesn't have any.

src/mappings/spacewalk.ts
gianfra-t: We get it from the `vaultId` yes

ebma:
@gianfra-t I pushed a commit with some refactoring. It was easier to already start the refactoring than describing it in text. I refactored the code related to the issue requests so that for all relevant events, the indexer just checks and compares it with the on-chain data. This way, the indexer will always have the accurate on-chain data (and we can also keep extra data like the timestamps around). I adjusted the schema of the issue request to match all fields that are available on the generated type. This also adds the debatable `asset` field but since we already have a function to parse it, we can just add it to that entity I guess. 
Let me know what you think. We can also revert that last commit if you have strong opinions against using this approach. It still needs more work, also applying the same changes to the redeem handling of course but I didn't want to do it if we don't want to go forth with this.

gianfra-t:
Ah! I see now. Thanks @ebma this seems way simpler. I had just finished solving an issue which was making the db panic when storing the redeem request (a typo, quite hard to find).

I will check if this approach also solved it, and if so, fine by me!

gianfra-t:
@ebma I finished the refactor with your new approach, the squids should be live now. I tested against a few events but not comprehensively all events.

TorstenStueber (approved):
Nice changes! Definitely simpler that way.

src/mappings/spacewalk.ts
```diff
@@ -0,0 +1,188 @@
...
+            storageIssue.stellarAddress
+        ),
+        griefingCollateral: BigInt(storageIssue.griefingCollateral),
+        status: getIssueRequestStatus(storageIssue.status),
+    })
+
+    await ctx.store.save(issueRequest)
+}
+
+export async function handleIssueRequest(ctx: EventHandlerContext) {
```
ebma: Nit: Let's rename this and `handleRedeemRequest` to `handleXXRequestCreated`.

src/processor.ts
```diff
@@ -128,6 +137,14 @@ const processor = new SubstrateBatchProcessor()
...
             'Contracts.ContractEmitted',
             'Contracts.Instantiated',
+            // Spacewalk
+            'Issue.RequestIssue',
+            'Issue.ExecuteIssue',
+            'Issue.CancelIssue',
+            'Issue.IssueAmountChange',
+            'Redeem.RequestRedeem',
+            'Redeem.ExecuteRedeem',
+            'Redeem.CancelRedeem',
```
ebma, line 148: Nit: let's also handle the other events that can modify redeem requests.

```suggestion
            'Issue.RequestIssue',
            'Issue.ExecuteIssue',
            'Issue.CancelIssue',
            'Issue.IssueAmountChange',
            'Redeem.RequestRedeem',
            'Redeem.ExecuteRedeem',
            'Redeem.CancelRedeem',
            'Redeem.LiquidationRedeem',
            'Redeem.MintTokensForReimbursedRedeem'
```

ebma (commented):
I like it 🙏  just pointed out two minor things we should still add.

src/processor.ts
gianfra-t: Sure. I didn't think those could modify the initial request state, but now that we are just tracking the instance in the storage there is no effort.

src/mappings/spacewalk.ts
gianfra-t: You mean have only one for any request creation? But then we have to switch based on the event and is more or less the same right?

gianfra-t: Ah... You mean only add `....Created`. Sure.

src/processor.ts
gianfra-t: I actually added both events into the types. But I think we cannot handle `LiquidationRedeem`, since it is not part of any particular redeem (as such it doesn't return any `redeemId`), and the vaults entity we keep track just tracks the id and currency pair and not liquidation-related fields.

ebma: Oh right, thanks for pointing it out. Since the liquidation redeem happens against the liquidation vault that is entirely managed by Spacewalk, it all happens in one go and not against a specific decentralized vault. We don't need to consider that event then, sorry for the confusion.

ebma (approved):

PR #77, opened 2024-10-22
Bind contract objects to appropriate blocks

### LINKED ISSUE #76: Bugfix: read historical state of contracts
When processing contract events, usually the current state of the contract as stored in the database is to be updated with the current actual state of the contract.

This happens by reading the current contract state. However, the indexer always reads the latest contract state and not the state of the contract at the block where the event happened.

This is usually not a problem as our apps read and require the latest contract state anyway. However, when executing scripts to process historical events (e.g., to determine rewards), then the data is corrupt.

## TODO
Read historical event data at the block of the respective event.

TorstenStueber:
Closes #76


ebma (approved):
Nice find that we can supply the block hash. Looks good to me 👍

PR #83, opened 2025-11-13
Fix indexer getting stuck randomly due to missing parent `call` entry.

gianfra-t:
Removes the `parent` field for the `Call` model. This prevents the error where a call would point to a missing parent entry, breaking the synchronisation. 

Since we do not use the `parent` field for any analysis, this doesn't affect the operation of the indexer.


ebma (approved):
Let's go 😎 I see that we currently have v29 on amplitude and foucoco. So let's bump the version for all three networks to v30 and deploy squids with that number to make it coherent again.

gianfra-t:
@ebma all versions are at v30 now (and deployed).

ebma (approved):

